PR #3, opened 2020-11-11
Initial producer/consumer exchanging

ppatierno:
This PR adds a very initial version of producer and consumer just exchanging messages.
They are defined as "services" (a new one about admin and topic creation/handling will be added) and started asynchronously by the main application.
There is a first configuration through env vars and the main application can stop producer/consumer only via SIGINT or SIGKILL signals.
The payload is not using the agreed message scheme yet.
There are no metrics right now.


internal/config/canary_config.go
```diff
@@ -0,0 +1,87 @@
...
+	ProducerClientIDEnvVar = "PRODUCER_CLIENT_ID"
+	ConsumerGroupIDEnvVar  = "CONSUMER_GROUP_ID"
+	TLSEnabledEnvVar       = "TLS_ENABLED"
+)
+
+// default values for environment variables
+const (
+	BootstrapServersDefault = "localhost:9092"
+	TopicDefault            = "strimzi-canary"
+	SendRateDefault         = 1000
```
scholzj, line 30: Is this a send rate as suggested in the name? Or is this interval between messages?

ppatierno: I got your point, it's not a send rate, it's a delay between messages. I will rename it ;-)

scholzj (approved):

robshelly (approved):

internal/services/consumer.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	log.Printf("Consumer closed")
+}
+
+// struct defining the handler for the consuming Sarama method
+type consumerGroupHandler struct {
+}
+
+func (cgh *consumerGroupHandler) Setup(sarama.ConsumerGroupSession) error {
+	log.Printf("Consumer group handler setup\n")
```
jeremyary, line 63: Given Peter's feedback about noise levels (and just a general eye towards growing the project), I'm wondering if long-term we shouldn't consider a logging framework to allow some differentiation between log levels. Not a show-stopper for the PR, but we should probably capture that somewhere to revisit. I can add a JIRA if you agree.

internal/services/producer.go
```diff
@@ -0,0 +1,67 @@
...
+	go func() {
+		for {
+			msg := &sarama.ProducerMessage{
+				Topic: p.config.Topic,
+				Value: sarama.StringEncoder(payload),
+			}
+			log.Printf("Sending message: value=%s\n", msg.Value)
+			partition, offset, err := p.producer.SendMessage(msg)
+			if err != nil {
+				log.Printf("Erros sending message: %v\n", err)
```
jeremyary, line 50: Same concerns for Producer panic/exits akin to Consumer, but this may be another instance where we care about the error and why it's occurring at a higher level.

jeremyary (commented):
Willing to approve pending your own interpretation of changes desired - I think most of what I've touched on could be personal preference or items to re-visit in follow-up efforts rather than gatekeeping here. See what you think regarding feedback.

cmd/main.go
```diff
@@ -0,0 +1,38 @@
...
+	producer := services.NewProducer(config)
+	producer.Start()
+
+	consumer := services.NewConsumer(config)
+	consumer.Start()
+
+	select {
+	case sig := <-signals:
+		log.Printf("Got signal: %v\n", sig)
+	}
```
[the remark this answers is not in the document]
ppatierno: I am already using goling (even inside VSCode) and it's not complaining. Anyway the `select` should have other condition to exit, so let's see it as a placeholder for now.

internal/config/canary_config.go
```diff
@@ -0,0 +1,87 @@
...
+	BootstrapServersEnvVar = "KAFKA_BOOTSTRAP_SERVERS"
+	TopicEnvVar            = "TOPIC"
+	DelayEnvVar            = "DELAY_MS"
+	ProducerClientIDEnvVar = "PRODUCER_CLIENT_ID"
+	ConsumerGroupIDEnvVar  = "CONSUMER_GROUP_ID"
+	TLSEnabledEnvVar       = "TLS_ENABLED"
+)
+
+// default values for environment variables
+const (
```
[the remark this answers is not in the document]
ppatierno: That makes sense to me. I thought that this was a way of grouping related constants, but I would agree that a comment is enough.

internal/services/consumer.go
```diff
@@ -0,0 +1,78 @@
...
+	config        *config.CanaryConfig
+	consumerGroup sarama.ConsumerGroup
+}
+
+func NewConsumer(config *config.CanaryConfig) Service {
+	// TODO: add specific consumer configuration
+	consumerGroup, err := sarama.NewConsumerGroup([]string{config.BootstrapServers}, config.ConsumerGroupID, nil)
+	if err != nil {
+		log.Printf("Error creating the Sarama consumer: %v", err)
+		panic(err)
```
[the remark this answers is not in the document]
ppatierno: Completely agree. This PR is just about starting up the codebase having some stuff working to exchange messages. We should track this as an issue.

internal/services/consumer.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}()
+}
+
+func (c *Consumer) Stop() {
+	log.Printf("Stopping consumer")
+	err := c.consumerGroup.Close()
+	if err != nil {
+		log.Printf("Error closing the Sarama consumer: %v", err)
+		os.Exit(1)
```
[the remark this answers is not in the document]
ppatierno: Ditto as above.

internal/services/consumer.go
ppatierno: Totally agree. We need something better than just log. As already mentioned just a PR to start up something working.

internal/services/producer.go
ppatierno: Ditto as above.

jeremyary (approved):

PR #6, opened 2020-11-11
Added JSON message schema

ppatierno:
This trivial PR added the JSOM message schema as defined by the proposal instead of the current fixed "payload".
It doesn't add the proper message index yet that could be different for each partition.
This will be done in future PR.

### DIFF
--- a/internal/services/consumer.go
+++ b/internal/services/consumer.go
@@ -8,6 +8,7 @@ package services
 
 import (
 	"context"
+	"encoding/json"
 	"log"
 	"os"
 
@@ -71,8 +72,15 @@ func (cgh *consumerGroupHandler) Cleanup(sarama.ConsumerGroupSession) error {
 
 func (cgh *consumerGroupHandler) ConsumeClaim(session sarama.ConsumerGroupSession, claim sarama.ConsumerGroupClaim) error {
 	for message := range claim.Messages() {
-		log.Printf("Message received: value=%s, partition=%d, offset=%d", string(message.Value), message.Partition, message.Offset)
+		cm := jsonToCanaryMessage(message.Value)
+		log.Printf("Message received: value=%+v, partition=%d, offset=%d", cm, message.Partition, message.Offset)
 		session.MarkMessage(message, "")
 	}
 	return nil
+}
+
+func jsonToCanaryMessage(value []byte) CanaryMessage {
+	var cm CanaryMessage
+	json.Unmarshal(value, &cm)
+	return cm
 }
--- a/internal/services/consumer.go
+++ b/internal/services/consumer.go
@@ -8,7 +8,6 @@ package services
 
 import (
 	"context"
-	"encoding/json"
 	"log"
 	"os"
 
@@ -72,15 +71,9 @@ func (cgh *consumerGroupHandler) Cleanup(sarama.ConsumerGroupSession) error {
 
 func (cgh *consumerGroupHandler) ConsumeClaim(session sarama.ConsumerGroupSession, claim sarama.ConsumerGroupClaim) error {
 	for message := range claim.Messages() {
-		cm := jsonToCanaryMessage(message.Value)
+		cm := NewCanaryMessage(message.Value)
 		log.Printf("Message received: value=%+v, partition=%d, offset=%d", cm, message.Partition, message.Offset)
 		session.MarkMessage(message, "")
 	}
 	return nil
-}
-
-func jsonToCanaryMessage(value []byte) CanaryMessage {
-	var cm CanaryMessage
-	json.Unmarshal(value, &cm)
-	return cm
 }
--- a/internal/services/producer.go
+++ b/internal/services/producer.go
@@ -7,7 +7,6 @@
 package services
 
 import (
-	"encoding/json"
 	"log"
 	"os"
 	"time"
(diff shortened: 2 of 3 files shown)


scholzj (approved):

internal/services/producer.go
```diff
@@ -36,13 +38,14 @@ func NewProducer(config *config.CanaryConfig) Service {
 
 func (p *Producer) Start() {
 	log.Printf("Starting producer")
-	// TODO: define the actual payload based on the agreed JSON schema
-	payload := "payload"
 	go func() {
 		for {
+			timestamp := time.Now().UnixNano() / 1000000 // timestamp in milliseconds
```
jeremyary, line 43: We should probably UTC this to prevent timezone complications down the road. `Now()` is going to yield server's local time.

internal/services/consumer.go
```diff
@@ -71,8 +72,15 @@ func (cgh *consumerGroupHandler) Cleanup(sarama.ConsumerGroupSession) error {
...
 		session.MarkMessage(message, "")
 	}
 	return nil
 }
+
+func jsonToCanaryMessage(value []byte) CanaryMessage {
+	var cm CanaryMessage
+	json.Unmarshal(value, &cm)
+	return cm
+}
```
jeremyary, line 86: How would you feel about moving this conversion out of the consumer and into canary_message.go? Combined with the comment below, it would gather all our conversions into one place.
```
func NewCanaryMessage([]byte) CanaryMessage
```

internal/services/producer.go
```diff
@@ -65,3 +68,13 @@ func (p *Producer) Stop() {
...
+
+func canaryMessageValueToJson(producerID string, index int, timestamp int64) string {
+	m := CanaryMessage{
+		ProducerID: producerID,
+		MessageID:  index,
+		Timestamp:  timestamp,
+	}
+	json, _ := json.Marshal(m)
+	return string(json)
+}
```
jeremyary, line 80: I feel like there's 2 operations happening here. WDYT about splitting it up as creation vs. conversion?
```
func (p *Producer) NewCanaryMessage() CanaryMessage {
    return CanaryMessage{
		ProducerID: p.config.ProducerClientID,
		MessageID:  p.index,
		Timestamp:  time.Now().UTC().UnixNano() / int64(time.Millisecond)
	}
}
```
and on CanaryMessage struct, add for conversion:
```
func (cm CanaryMessage) Json() string
```

internal/services/consumer.go
```diff
@@ -71,8 +72,15 @@ func (cgh *consumerGroupHandler) Cleanup(sarama.ConsumerGroupSession) error {
...
+		cm := jsonToCanaryMessage(message.Value)
+		log.Printf("Message received: value=%+v, partition=%d, offset=%d", cm, message.Partition, message.Offset)
 		session.MarkMessage(message, "")
 	}
 	return nil
 }
+
+func jsonToCanaryMessage(value []byte) CanaryMessage {
+	var cm CanaryMessage
+	json.Unmarshal(value, &cm)
```
jeremyary, line 84: any use case where we care about error handling the unmarshalling?

internal/services/producer.go
```diff
@@ -65,3 +68,13 @@ func (p *Producer) Stop() {
...
 	log.Printf("Producer closed")
 }
+
+func canaryMessageValueToJson(producerID string, index int, timestamp int64) string {
+	m := CanaryMessage{
+		ProducerID: producerID,
+		MessageID:  index,
+		Timestamp:  timestamp,
+	}
+	json, _ := json.Marshal(m)
```
jeremyary, line 78: is it safe for us to ignore the error here?

jeremyary (changes requested):

internal/services/producer.go
tombentley: But 

> UnixNano returns t as a Unix time, the number of nanoseconds elapsed since January 1, 1970 UTC

So things like DST changes shouldn't affect the value of `timestamp` right?

ppatierno: Yes, indeed. Were you thinking about something different @jeremyary ?

robshelly (approved):

internal/services/producer.go
jeremyary: No, Tom's right there, I glazed over the epoch implication and didn't look up the method. Disregard :)

internal/services/producer.go
ppatierno: I agree but I don't think that the canary message creation needs to be a method provided by a `Producer` instance and even publically available. It's not called outside. I would just have a `createCanaryMessage` or just leaving the code where it is right now. WDYT? 
For sure having the `Json()` method on the `CanaryMessage` struct makes more sense.

ppatierno: I changed my mind it would make sense as a kind of private method on `Producer` struct. :-)

internal/services/consumer.go
ppatierno: we are actually generating our own message, so it should be a valid JSON at the source/producer.

internal/services/producer.go
ppatierno: we are building the message, so we would expect valid values there but anyway we can think more about that adding some more error handling later.

jeremyary (approved):
lgtm!

PR #8, opened 2020-11-17
Added canary topic creation

ppatierno:
This PR adds a topic manager service that:

* creates the topic if it doesn't exist on startup, assigning replicas directly
* starts a reconcile loop to check periodically the status of the cluster to check if it was scaled up or down and the topic needs some rearrangement (adding partitions) or signaling to the producer about the "orphan" partitions (where messages should not be sent).

This PR doesn't include the logic for handling the cluster scale up and down yet; it will be added in a future PR.


internal/services/topic_manager.go
```diff
@@ -0,0 +1,145 @@
...
+		// canary topic already exists, check replicas assignments
+		log.Printf("The canary topic %s already exists\n", metadata[0].Name)
+
+		if len(metadata[0].Partitions) < len(brokers) {
+			log.Printf("Cluster scaled up, need to add partitions")
+			// TODO: logic for adding partitions and assigning them
+		} else if len(metadata[0].Partitions) > len(brokers) {
+			log.Printf("Cluster scaled down, need to declare 'orphan' partitions")
+			// TODO: need to track "orphan" partitions to avoid producer sending to them
+		}
```
scholzj, line 114: Just with regards to the TODOs ... this might also need to deal with the replication factor. Not to be implemented here, but maybe you can add it to the comments so that it is not forgotten?

scholzj (approved):

robshelly (approved):

internal/services/topic_manager.go
ppatierno: do you mean changing the replication factor accordingly and the minISR as well?

scholzj: Yes, for scale down / up you will need to tackle it as well.

ppatierno: My plan is to factor out this part of the topic creation:

```go
        partitions := brokersNumber
	replicationFactor := math.Min(float64(brokersNumber), 3)
	minISR := math.Max(1, replicationFactor-1)

	assignments := make(map[int32][]int32, partitions)
	for p := 0; p < partitions; p++ {
		assignments[int32(p)] = make([]int32, int(replicationFactor))
		k := p
		for r := 0; r < int(replicationFactor); r++ {
			assignments[int32(p)][r] = int32(k % int(replicationFactor))
			k++
		}
	}
```

it should be the piece tackling this because it's going to reassign partitions and replicas on brokers.
With scale up it should get a full replicas assignments, while when scaling down it should leave out some partitions/replicas but I would check how Kafka brokers will behaves on assignment in that case.
I mean for example, providing a reassignment for only 3 partitions out of 5, because the cluster is now 3 and not 5 anymore so I don't care where the other partitions will be or we should anyway reassign properly but leaving them "orphan" from a production pov. I will think more on this.

scholzj: You will need to schedule them somewhere even if you don' use them. Anyway, my main point was to add it to the TODO comment. Not necessarily implement it here.

ppatierno: Yeah sure.

jeremyary (approved):
lgtm!

PR #11, opened 2020-11-20
Refactoring with a manager driving topic and producer services

ppatierno:
This PR does a refactoring of the canary tool shape.
Mainly now we have:

* a canary manager worker driving the topic management and the producer services sending when the topic is in the "right" shape (the right number of partitions and replicas, and the right leaders on brokers)
* a consumer worker running for getting the messages

The overall idea is the following:

* The consumer is running on its own for getting the messages
* The canary manager has its own “reconcile” loop with a period that defines how often we want to send messages (i.e. every 1 sec, 5 secs, …). Every reconcile it checks the cluster topology, so the number of brokers, and:
   - Creating the topic if it doesn’t exist and assigning replicas in order to have each broker as the leader for one partition or ….
… getting the current topic replicas assignments
   - If the assignments are what it expects and the leaders as well, it triggers the producer to send messages to the specific 
partitions 
   - If the assignments are not what we expect or leaders are wrong, it does a reassignment and force leader election first then triggers the producer to send the messages

Pros:
* Every time before sending a message we have more updated information about the cluster topology, the assignments are right and the leaders as well
Cons:
* More traffic for getting cluster topology and topic metadata quite often but it should not impact the Kafka cluster too much

What's missing in this PR:

* getting current assignments if topic exists and compare to the desired one.Currently, a reassignment request is done anyway. Maybe it works ok, the broker should do nothing, instead of getting assignments, compare and "maybe" request another one. Wdyt @tombentley ?
* forcing the leader election (when scale down and scale up again, so existing partitions have to come back to their original leader). It's not available in Sarama so it means contributing to implement the Elect Leader API (in the Kafka protocol).


ppatierno:
I have just updated the PR with these changes:

* using the Sarama `client` which provides a common connection "handler" to both producer and consumer and also provides the possibility to force metadata refresh for the producer
* due to how Sarama works and reported here (https://github.com/Shopify/sarama/issues/1839), refreshing metadata doesn't really work for the consumer which needs to close a session and rejoining the group. I added this logic in the canary as a refreshing mechanism. 

@tombentley @robshelly can you review now please?

internal/services/consumer.go
```diff
@@ -15,39 +15,66 @@ import (
...
 	}
-	return &c
+	return &cs
 }
 
-func (c *Consumer) Start() {
+func (cs *consumerService) Consume() {
 	cgh := &consumerGroupHandler{}
-	ctx := context.Background()
+	// creating new context with cancellation, for exiting Consume when metadata refresh is needed
```
jeremyary, line 47: what does triggering a metadata refresh do? Is it to re-fetch broker info?

internal/services/topic.go
```diff
@@ -0,0 +1,171 @@
...
+	client       sarama.Client
+	admin        sarama.ClusterAdmin
+}
+
+func NewTopicService(canaryConfig *config.CanaryConfig, client sarama.Client) TopicService {
+	/*
+		adminConfig := sarama.NewConfig()
+		adminConfig.Version = sarama.V2_6_0_0
+		admin, err := sarama.NewClusterAdmin([]string{canaryConfig.BootstrapServers}, adminConfig)
+	*/
```
jeremyary, line 35: is commented code left here as a reminder of what config init looks like or is it an artifact of the refactor that can be removed?

ppatierno: It's a leftover ... removing it, thanks!

internal/services/consumer.go
```diff
@@ -15,39 +15,66 @@ import (
 	"github.com/strimzi/strimzi-canary/internal/config"
 )
 
-type Consumer struct {
-	config        *config.CanaryConfig
+type ConsumerService interface {
+	Consume()
+	Refresh()
+	Close()
+}
```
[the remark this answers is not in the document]
ppatierno: Actually, there is not so much gain in this case.
Maybe we are going to add a new Producer for reaching the Kafka cluster from outside the K8S and that interface could be useful; maybe not.
Do you think we should remove them and readding them only in case we find them really useful?

internal/services/topic.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	metadata, err := ts.admin.DescribeTopics([]string{ts.canaryConfig.Topic})
+	if err != nil {
+		log.Printf("Error retrieving metadata for topic %s: %v", ts.canaryConfig.Topic, err)
+		return len(brokers), nil, refresh, err
+	}
+
+	var assignments map[int32][]int32
+	if metadata[0].Err == sarama.ErrUnknownTopicOrPartition {
```
tombentley, line 61: Can we factor out a var for `metadata[0]` since you used it multiple times?

internal/services/topic.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	ts := topicService{
+		canaryConfig: canaryConfig,
+		client:       client,
+		admin:        admin,
+	}
+	return &ts
+}
+
+func (ts *topicService) Reconcile() (int, map[int32][]int32, bool, error) {
```
tombentley, line 44: Some doc on the return tuple would be good. Or perhaps the return value should be a type, rather than a 4-tuple?

internal/services/topic.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+		}
+	}
+	log.Printf("Elect leader = %t\n", electLeader)
+}
+
+func (ts *topicService) assignments(currentPartitions int, brokersNumber int) (map[int32][]int32, int) {
+	partitions := math.Max(float64(currentPartitions), float64(brokersNumber))
+	replicationFactor := math.Min(float64(brokersNumber), 3)
+	minISR := math.Max(1, replicationFactor-1)
```
tombentley, line 153: It looks like we're using floating point just so we can use `math.Max`. It would be cleaner imho you just write your own `max` function and stick with integer arithmetic.

internal/services/topic.go
```diff
@@ -0,0 +1,166 @@
...
+import (
+	"log"
+	"math"
+	"os"
+	"strconv"
+
+	"github.com/Shopify/sarama"
+	"github.com/strimzi/strimzi-canary/internal/config"
+)
+
```
tombentley, line 18: In general in this file I feel that a little more documentation would go along way. There are a lot of ints flying around and I don't know whether they're numbers of brokers, broker ids, numbers of replicas or replication factors.

cmd/main.go
```diff
@@ -10,33 +10,44 @@ import (
...
+	canaryManager.Stop()
 	log.Printf("Strimzi canary stopped")
 }
+
+func newClient(canaryConfig *config.CanaryConfig) sarama.Client {
+	config := sarama.NewConfig()
+	config.Version = sarama.V2_6_0_0
+	// set manual partitioner in order to specify the destination partition on sending
+	config.Producer.Partitioner = sarama.NewManualPartitioner
+	config.Producer.Return.Successes = true
```
tombentley, line 49: You should set the `config.ClientId`

cmd/main.go
```diff
@@ -10,33 +10,44 @@ import (
...
-	topicManager.Stop()
-	producer.Stop()
-	consumer.Stop()
+	canaryManager.Stop()
 	log.Printf("Strimzi canary stopped")
 }
+
+func newClient(canaryConfig *config.CanaryConfig) sarama.Client {
+	config := sarama.NewConfig()
+	config.Version = sarama.V2_6_0_0
```
tombentley, line 46: Any reason not to make this a config?

robshelly (approved):
:+1: Looks good!

tombentley (commented):
Reviewing this as a non-golang programmer (well, I once read half a book about it...) some of these comments might not make sense for go, or might not be idiomatic. So feel free to ignore those (and better if you can educate me, but that's optional).

1. It would be good to see some tests. Things like `topicService.assignments` aren't entirely trivial. 

2. Some comments describing what some of those methods are supposed to do. E.g. Under what circumstances would `topicService.Reconcile` return an error and is it true that there will be as many assignments as brokers in the return value?

internal/services/topic.go
ppatierno: Yeah, that's one of the problems with Go where you don't have an "overload" thing but you can just play around with functions with different names. I will write my own Max, anyway, we are going to use just integers.

cmd/main.go
ppatierno: The interesting thing is that it's just one field but the producer and consumer share this `Client` Go struct so both will have same clientid. I used it for having the refresh metadata available on demand. For now I can set just to a more generic `strimzi-canary`; later I will check if we should use two different `Client` instances for producer and consumer.

internal/services/consumer.go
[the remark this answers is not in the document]
jeremyary: IMHO, remove for now and re-add when we need to - I suspect they may come back soon enough around testing, but if we're not sure, I think it's good to remove it.

internal/services/topic.go
ppatierno: Maybe just a three fields type, leaving the error out because it's mostly a Golang pattern returning a result and eventually an error in a tuple fashion.

cmd/main.go
ppatierno: Yes it could be, I will open an issue to track this for now.

internal/services/topic.go
ppatierno: Yes, we need it, let me add it in a different PR. Opening an issue for tracking for now.

jeremyary (approved):

cmd/main.go
tombentley: I think it's fine for producer and client to have the same client id. But it should be configurable, if you're going to run multiple canaries, so you can distinguish them.

ppatierno: @tombentley agree, just pushed the change.

tombentley (approved):

Issue #10, opened 2020-11-17
Build Canary image


robshelly:
A Dockerfile should be added to allow building the image

ppatierno:
More than just a Dockerfile, the entire building pipeline should be developed.

PR #15, opened 2020-12-02
Added Dockerfile and Makefile to build image

robshelly:
Dockerfile and Makefile added to allow building and pushing the image. Makefile includes `build`, `tag` and `push` targets.

However as there has not been an initial release of the Strimzi Canary yet there are no default values for the image registry and organisation in Makefile. They must be specified as arguments in order to tag and push the image. 

Addresses #10

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM centos:7
+
+ARG version
+ENV VERSION=${version}
+
+ADD target/main /
+
+CMD ["/main"]
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,33 @@
+IMAGE ?= strimzi-canary
+VERSION ?= latest
+TAG ?= latest
+
+
+.PHONY: help dobuild build tag push all
+
+help:
+	@echo "Makefile arguments:"
+	@echo "REGISTRY: Image registry"
+	@echo "ORG: Image organisation"
+	@echo "Makefile commands:"
+	@echo "build"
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM golang:1.14-alpine
+
+ARG version
+ENV VERSION=${version}}
```
ppatierno, line 4: is this double `}` a typo?

Makefile
```diff
@@ -0,0 +1,30 @@
+IMAGE					?= strimzi-canary
+VERSION				?= latest
+TAG           ?= latest
```
ppatierno, line 3: why all these spaces?

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN mkdir /app
+ADD . /app
+WORKDIR /app
+
+COPY go.mod go.sum ./
+
+RUN go mod download
+
+RUN go build -o main ./cmd/
```
tombentley, line 14: Is it normal to build the executable in the container like this, thus the resulting image containing the go toolchain?

scholzj (commented):
I do not think we want to use the build image as runtime image. I think the runtime image should be based on CentOS:7 as our other images. 

I think it would be also great to follow similar Makefile schema as we have in other projects as closely as possible. That will make it easier to use and to develop.

ppatierno:
Going to close this replaced by #40

PR #17, opened 2020-12-16
Fix replica assignment invalid

### LINKED ISSUE #16: "Replica assignment is invalid." sometimes seen when starting the canary.
The canary is manually co-deployed with strimzi.  Sometimes, when the canary starts up, the following message repeats in the logs.  

Restarting the canary doesn't clear the condition.

Also I notice that there's no metric counting these failures, so when this condition occurs, a team's SREs would be blind to the condition.

```
2020/12/11 12:36:51 The canary topic strimzi-canary already exists
2020/12/11 12:36:51 assignments = map[0:[0 1 2] 1:[1 2 0] 2:[2 0 1]], minISR = 2
2020/12/11 12:36:51 [[0 1 2] [1 2 0] [2 0 1]]
2020/12/11 12:36:51 Error altering topic strimzi-canary: kafka server: Replica assignment is invalid. - Inconsistent replication factor between partitions, partition 0 has 2 while partitions [2] have replication factors [3], respectively.
2020/12/11 12:36:51 ... reconcile done
```

```
curl --silent http://10.129.2.47:8080/metrics | grep strimzi_
# HELP strimzi_canary_records_produced_latency Records produced latency in milliseconds
# TYPE strimzi_canary_records_produced_latency histogram
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="0",le="100"} 0
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="0",le="200"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="0",le="400"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="0",le="800"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="0",le="1600"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="0",le="+Inf"} 1
strimzi_canary_records_produced_latency_sum{clientid="strimzi-canary-client",partition="0"} 182
strimzi_canary_records_produced_latency_count{clientid="strimzi-canary-client",partition="0"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="1",le="100"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="1",le="200"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="1",le="400"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="1",le="800"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="1",le="1600"} 1
strimzi_canary_records_produced_latency_bucket{clientid="strimzi-canary-client",partition="1",le="+Inf"} 1
strimzi_canary_records_produced_latency_sum{clientid="strimzi-canary-client",partition="1"} 3
strimzi_canary_records_produced_latency_count{clientid="strimzi-canary-client",partition="1"} 1
# HELP strimzi_canary_records_produced_total The total number of records produced
# TYPE strimzi_canary_records_produced_total counter
strimzi_canary_records_produced_total{clientid="strimzi-canary-client",partition="0"} 1
strimzi_canary_records_produced_total{clientid="strimzi-canary-client",partition="1"} 1
```

ppatierno:
This PR fixes #16.
As described in the committed code, this happens in the following scenario.

When the replication factor is less than 3 because brokers are not 3 yet (see `replicationFactor := min(brokersNumber, 3)`),
it's not possible to create the new partitions directly with a replication factor higher than the current ones.
So first alter the assignment of current partitions with new replicas (higher replication factor).
It's not needed anymore when brokers are 3 or more because the replication factor (as 3 max) cannot be higher than that.

The PR does what described even checking that there is no ongoing reassignment ongoing before doing the new partition creation.

### DIFF
--- a/internal/services/topic.go
+++ b/internal/services/topic.go
@@ -148,15 +148,14 @@ func (ts *TopicService) alterTopic(currentPartitions int, brokersNumber int) (ma
 		// when replication factor is less than 3 because brokers are not 3 yet (see replicationFactor := min(brokersNumber, 3)),
 		// it's not possible to create the new partitions directly with a replication factor higher than the current ones.
 		// So first alter the assignment of current partitions with new replicas (higher replication factor)
-		// fixes: https://github.com/strimzi/strimzi-canary/issues/16
-		err = ts.alterAssignments(ts.canaryConfig.Topic, ass[:currentPartitions])
+		err = ts.alterAssignments(ass[:currentPartitions])
 		if err == nil {
 			// passing the assigments just for the partitions that needs to be created
 			err = ts.admin.CreatePartitions(ts.canaryConfig.Topic, int32(brokersNumber), ass[currentPartitions:], false)
 		}
 	} else {
 		// more or equals partitions than brokers, just need reassignment
-		err = ts.alterAssignments(ts.canaryConfig.Topic, ass[:currentPartitions])
+		err = ts.alterAssignments(ass[:currentPartitions])
 	}
 	return assignments, err
 }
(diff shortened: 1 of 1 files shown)


ppatierno:
@k-wall FYI

internal/services/topic.go
```diff
@@ -144,11 +145,18 @@ func (ts *TopicService) alterTopic(currentPartitions int, brokersNumber int) (ma
 	var err error
 	// less partitions than brokers (scale up)
 	if currentPartitions < brokersNumber {
-		// passing the assigments just for the partitions that needs to be created
-		err = ts.admin.CreatePartitions(ts.canaryConfig.Topic, int32(brokersNumber), ass[currentPartitions:], false)
+		// when replication factor is less than 3 because brokers are not 3 yet (see replicationFactor := min(brokersNumber, 3)),
+		// it's not possible to create the new partitions directly with a replication factor higher than the current ones.
+		// So first alter the assignment of current partitions with new replicas (higher replication factor)
+		// fixes: https://github.com/strimzi/strimzi-canary/issues/16
```
tombentley, line 151: I don't think we need comments in the code about (what will become) fixed issues. Just describe this case if necessary.

internal/services/topic.go
```diff
@@ -183,6 +191,36 @@ func (ts *TopicService) assignments(currentPartitions int, brokersNumber int) (m
 	return assignments, int(minISR)
 }
 
+func (ts *TopicService) alterAssignments(topic string, assignments [][]int32) error {
```
tombentley, line 194: Why make it take the topic as a parameter when it's available as `ts.canaryConfig.Topic`?

Also a line of doc about what this function is trying to achieve wouldn't hurt.

internal/services/topic.go
```diff
@@ -183,6 +191,36 @@ func (ts *TopicService) assignments(currentPartitions int, brokersNumber int) (m
...
+	}
+	// loop for checking that there is no ongoing reassignments
+	for {
+		ongoing := false
+		reassignments, err := ts.admin.ListPartitionReassignments(topic, partitions)
+		if err != nil {
+			return nil
+		}
+		// on each partition of the topic shouldn't be adding or removing replicas ongoing
+		for _, v := range reassignments[topic] {
```
tombentley, line 212: A better variable name than `v`?

internal/services/topic.go
```diff
@@ -183,6 +191,36 @@ func (ts *TopicService) assignments(currentPartitions int, brokersNumber int) (m
...
 }
 
+func (ts *TopicService) alterAssignments(topic string, assignments [][]int32) error {
+	err := ts.admin.AlterPartitionReassignments(topic, assignments)
+	if err != nil {
+		return err
+	}
+
+	partitions := make([]int32, 0, len(assignments))
+	for k := range assignments {
```
tombentley, line 201: A better variable name than `k`? `assignment` perhaps?

tombentley (approved):
A few nits, but otherwise LGTM.

internal/services/topic.go
ppatierno: I used `reassignmentStatus`

internal/services/topic.go
ppatierno: I changed into `partitionID`

PR #20, opened 2020-12-17
Fixed consumer starting first and be subscribed before producing

### LINKED ISSUE #19: Consumer miss first message(s) on topic creation or after replicas reassignment
Currently, when the canary manager starts, it does a first reconcile which creates the topic (or check if it exists doing if needed, a replica reassignment) and then starts the producer, finally, it starts the consumer.
In this way, the first message(s) sent by the producer is/are not received by the consumer (that uses the default offset reset policy as "latest").
The best thing to do would be to allow the topic creation/check, starting the consumer first, and then producing messages in order to allow the consumer to be "listening" before messages come in. 
This happens, not only on topic creation but even when there is a Kafka cluster scale up/down so a replicas assignment is needed because the consumer has to refresh metadata and rejoin the group.

ppatierno:
This PR fixes #19 
In order to avoid first messages to be "lost" and not received, it starts the consumer first, waits for it to be subscribed to the partitions and then continue with producing messages.
This is done the first time at startup on topic creation or anyway on each reconcile when consumer and producer need to refresh metadata.

### DIFF
--- a/internal/services/consumer.go
+++ b/internal/services/consumer.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"os"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/Shopify/sarama"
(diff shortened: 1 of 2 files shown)


tombentley (approved):

ppatierno:
After talking offline, I am going to merge this one and working on adding a timeout on waiting for consumer starting up in a new PR.

PR #22, opened 2020-12-21
Added waiting with timeout for consumer to join group

### LINKED ISSUE #21: Timeout on consumer starting and metric on joining group errors
Currently, the consumer is started before the producer to avoid missing messages on starting up.
Anyway, it could be possible that the consumer is not able to start because of edge cases when the group coordinator cannot be elected.
An example could be the following:

To elect the group coordinator, you basically take the group id, hash it, modulo the #partitions of `__consumer_offsets` and that gives you a partition of  `__consumer_offsets`. The current leader of that partition is the group coordinator. So if there is no leader (e.g. the ISR is a singleton and that broker is down) then no group and no consumption.

In this scenario, we should report an error exposing a corresponding metric and keep retrying.
It would be useful having a timeout on waiting for the consumer to be ready (joined the group) and if the timeout is raised:

1. cancel the corresponding Sarama consumer context to avoid "internal" attempts to join the group
2. reporting the error about not able to join the group on time -> update a corresponding metric
3. retry creating the consumer and joining the group

ppatierno:
This PR fixes #21 
It introduces waiting with timeout for the consumer to start properly joining the consumer group before starting the producer for sending.
When timed out, it increases a corresponding metric counter; then it retries again after a delay.

### DIFF
--- a/internal/services/consumer.go
+++ b/internal/services/consumer.go
@@ -23,7 +23,7 @@ import (
 const (
 	// timeout on waiting the consumer to join the consumer group successfully
 	waitConsumeTimeout = 30 * time.Second
-	// delay betweet retries about the consumer to join the consumer group
+	// delay between retries for the consumer to join the consumer group
 	retryConsumeDelay = 5 * time.Second
 )
 
--- a/internal/services/consumer.go
+++ b/internal/services/consumer.go
@@ -20,6 +20,13 @@ import (
 	"github.com/strimzi/strimzi-canary/internal/config"
 )
 
+const (
+	// timeout on waiting the consumer to join the consumer group successfully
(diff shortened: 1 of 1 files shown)


internal/services/consumer.go
```diff
@@ -20,6 +20,13 @@ import (
 	"github.com/strimzi/strimzi-canary/internal/config"
 )
 
+const (
+	// timeout on waiting the consumer to join the consumer group successfully
+	waitConsumeTimeout = 30 * time.Second
+	// delay betweet retries about the consumer to join the consumer group
```
tombentley, line 26:
```suggestion
	// delay between retries for the consumer to join the consumer group
```

tombentley (approved):

PR #24, opened 2020-12-22
Added backoff for consumer retrying to join consumer group

### LINKED ISSUE #23: Add backoff mechanism for consumer to retry join group
Currently, if consumer is not able to join consumer group it keeps retrying with a fixed delay forever.
We should add a backoff mechanism in order to have different delays and specify a max number of attempts as well.

ppatierno:
This PR fixes #23 
It adds a backoff mechanism for the consumer to delay and retry on joining a consumer group when it times out.

### DIFF
--- a/internal/services/backoff.go
+++ b/internal/services/backoff.go
@@ -13,7 +13,6 @@ import (
 
 const (
 	ScaleDefault       = 200
-	BaseDefault        = 2
 	MaxAttemptsDefault = 6
 )
 
@@ -21,7 +20,6 @@ const (
 type Backoff struct {
 	maxAttempts int
 	scale       int64
-	base        int
 	attempt     int
 }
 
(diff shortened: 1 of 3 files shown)


internal/services/backoff.go
```diff
@@ -0,0 +1,74 @@
...
+package services
+
+import (
+	"fmt"
+	"time"
+)
+
+const (
+	ScaleDefault       = 200
+	BaseDefault        = 2
```
tombentley, line 16: We only really need to support a base of 2, and it would simplify the code quite a bit if we made that assumption.

internal/services/backoff.go
```diff
@@ -0,0 +1,74 @@
...
+			n--
+			if n <= 1 {
+				break
+			}
+			pow *= b.base
+		}
+	*/
+	for ; n > 1; n-- {
+		pow *= b.base
+	}
```
tombentley, line 72: With a base 2 assumption we could turn this loop into a bitshift.

tombentley (approved):
A minor comment, but otherwise LGTM.

tombentley (approved):

Issue #5, opened 2020-11-11
Adding proper error handling


ppatierno:
Currently, with most of the error, the canary tool just panics leaving the application to exit.
We should think about adding proper error handling where it makes sense, trying a retry or recovery logic, or even just exiting.

k-wall:
+1 with the idea of retry, but we need to make sure that an SRE observing the metrics have sufficient metrics to know that the canary is in trouble.  I'd suggest a monotomicly increasing error count metric.

ppatierno:
I totally agree with this, having an error metric with one or more labels specifying the kind of error. But do you agree that it can be addressed working on #5? It's more a wider problem than just this issue.
Maybe you can add this comment on #5 so that we can track the idea about a corresponding metric?

Issue #4, opened 2020-11-11
Using a proper logging framework


ppatierno:
The current default `log` package doesn't allow us to differentiate between logging levels making a lot of noise.
We should add a proper logging framework in order to log different operations with different logging levels like INFO, DEBUG, WARN, ERROR and so on.

PR #30, opened 2021-04-07
Adds configuration of expected brokers to avoid dynamic reassignment

### LINKED ISSUE #28: Configure canary to not do dynamic reassignment during periodic reconcile
I stumbled into a weird scenario when a Strimzi upgrade happens driving rolling updates of the ZooKeeper nodes and Kafka brokers to newer images version when the canary is running against this cluster.
What I noticed is that the last rolled Kafka broker has a 100% CPU usage for a long time (even more than a couple of hours) and then it comes back to normal usage later (with no intervention).
After digging into logs and some offline chats with @tombentley (who can provide more details), it seems that it's a replica fetcher thread looping and taking the CPU may be due to a Kafka bug but the main cause of that is the canary doing partitions reassignment when the overall cluster is rolling.
Of course, the canary tool doesn't know that Kafka brokers are disappearing and reappearing due to a rolling update but it just sees the cluster scaling down and up again so applying the partitions reassignment logic during these steps.

In order to mitigate this problem (until the root cause is identified and fixed in Kafka) it could be possible to configure the canary with the "expected" number of brokers that makes the cluster avoiding to do the reassignment; it's the kind of the opposite of the current behavior which is more dynamic by addressing scaling up/down.

On startup the canary could create the topic when the expected broker's number is what it sees on the cluster and even doing reassignment if there is a need due to a scale down/up; anyway it can happen only at startup not during periodic reconcile.
Of course, it means that if there is a scale up/down of the cluster, the canary tool needs to be restarted.

ppatierno:
This PR fixes #28 allowing to specify the expected number of brokers on the cluster, creating the topic and assigning partitions accordingly but not handling a dynamic reassignment in case of scale up/down. In this case, the canary has to be restarted and configured with the new expected number of brokers after the scale up/down.

### DIFF
--- a/internal/services/topic.go
+++ b/internal/services/topic.go
@@ -80,24 +80,41 @@ func (ts *TopicService) Reconcile() (TopicReconcileResult, error) {
 	topicMetadata := metadata[0]
 
 	if topicMetadata.Err == sarama.ErrUnknownTopicOrPartition {
+
 		// canary topic doesn't exist, going to create it
 		log.Printf("The canary topic %s doesn't exist\n", topicMetadata.Name)
-		if result.Assignments, err = ts.createTopic(len(brokers)); err != nil {
-			log.Printf("Error creating topic %s: %v", topicMetadata.Name, err)
-			return result, err
+		// topic is created if "dynamic" reassignment is enabled or the expected brokers are provided by the describe cluster
+		if ts.canaryConfig.ExpectedClusterSize == config.ExpectedClusterSizeDefault ||
+			ts.canaryConfig.ExpectedClusterSize == len(brokers) {
+
+			if result.Assignments, err = ts.createTopic(len(brokers)); err != nil {
+				log.Printf("Error creating topic %s: %v", topicMetadata.Name, err)
+				return result, err
+			}
+			log.Printf("The canary topic %s was created\n", topicMetadata.Name)
+		} else {
+			// not creating the topic and returning error to avoid starting producer/consumer
+			return result, topicMetadata.Err
 		}
-		log.Printf("The canary topic %s was created\n", topicMetadata.Name)
+
 	} else {
 		// canary topic already exists, check replicas assignments
 		log.Printf("The canary topic %s already exists\n", topicMetadata.Name)
 
-		result.RefreshMetadata = len(brokers) != len(topicMetadata.Partitions)
-		if result.Assignments, err = ts.alterTopic(len(topicMetadata.Partitions), len(brokers)); err != nil {
-			log.Printf("Error altering topic %s: %v", topicMetadata.Name, err)
-			return result, err
+		// topic partitions reassignment happens if "dynamic" reassignment is enabled
+		// or partitions are not equals to the expected brokers
+		if ts.canaryConfig.ExpectedClusterSize == config.ExpectedClusterSizeDefault ||
+			len(topicMetadata.Partitions) != ts.canaryConfig.ExpectedClusterSize {
+
+			result.RefreshMetadata = len(brokers) != len(topicMetadata.Partitions)
+			if result.Assignments, err = ts.alterTopic(len(topicMetadata.Partitions), len(brokers)); err != nil {
+				log.Printf("Error altering topic %s: %v", topicMetadata.Name, err)
+				return result, err
+			}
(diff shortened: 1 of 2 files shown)


internal/services/topic.go
```diff
@@ -80,24 +80,41 @@ func (ts *TopicService) Reconcile() (TopicReconcileResult, error) {
...
 
 	if topicMetadata.Err == sarama.ErrUnknownTopicOrPartition {
+
 		// canary topic doesn't exist, going to create it
 		log.Printf("The canary topic %s doesn't exist\n", topicMetadata.Name)
-		if result.Assignments, err = ts.createTopic(len(brokers)); err != nil {
-			log.Printf("Error creating topic %s: %v", topicMetadata.Name, err)
-			return result, err
+		// topic is created if "dynamic" reassignment is enabled or the expected brokers are provided by the describe cluster
+		if ts.canaryConfig.ExpectedClusterSize == config.ExpectedClusterSizeDefault ||
```
tombentley, line 87: Clearer to factor out some `isDynamicReassignmentEnabled()` function

internal/services/topic.go
```diff
@@ -80,24 +80,41 @@ func (ts *TopicService) Reconcile() (TopicReconcileResult, error) {
...
+			ts.canaryConfig.ExpectedClusterSize == len(brokers) {
+
+			if result.Assignments, err = ts.createTopic(len(brokers)); err != nil {
+				log.Printf("Error creating topic %s: %v", topicMetadata.Name, err)
+				return result, err
+			}
+			log.Printf("The canary topic %s was created\n", topicMetadata.Name)
+		} else {
+			// not creating the topic and returning error to avoid starting producer/consumer
+			return result, topicMetadata.Err
```
tombentley, line 97: So does this mean we crash loop until the cluster has the expected size? Should we log the reason we're returning here, since the caller won't be able to distinguish this from other errors, will it?

internal/services/topic.go
```diff
@@ -80,24 +80,41 @@ func (ts *TopicService) Reconcile() (TopicReconcileResult, error) {
...
 		log.Printf("The canary topic %s already exists\n", topicMetadata.Name)
 
-		result.RefreshMetadata = len(brokers) != len(topicMetadata.Partitions)
-		if result.Assignments, err = ts.alterTopic(len(topicMetadata.Partitions), len(brokers)); err != nil {
-			log.Printf("Error altering topic %s: %v", topicMetadata.Name, err)
-			return result, err
+		// topic partitions reassignment happens if "dynamic" reassignment is enabled
+		// or partitions are not equals to the expected brokers
+		if ts.canaryConfig.ExpectedClusterSize == config.ExpectedClusterSizeDefault ||
+			len(topicMetadata.Partitions) != ts.canaryConfig.ExpectedClusterSize {
```
tombentley, line 107: Just to be sure I understand, if we scale down then because Kafka doesn't support deleting partitions `len(partitions) > expectedSize`, so we reassign, but that's OK because we only run this at startup now, right?

internal/services/topic.go
ppatierno: Sad to say but yes, right now this is the way it works right now. There are two issues #4 and #5 to address better logging and error handling. Anyway, it makes sense to log a specific message in this case.

internal/services/topic.go
ppatierno: Reconcile is called at startup but even periodically.
Scaling down is obviously different from scale up. This condition works for scale up so reassignment is done only at startup even if the method is called more times but doesn't work for scale down where len(partitions) will be always higher than expected cluster size.
Good catch, have to fix this scenario.

ppatierno: What I am trying to do is actually trying to reuse the same Reconcile() method to avoid code duplication across code at startup time and periodic code for reconcile.

ppatierno: And actually the `len(topicMetadata.Partitions) != ts.canaryConfig.ExpectedClusterSize` doesn't work in case of scale down and scale up again when a reassignment should happen again but in this case won't.
I am simplifying the condition, I think that doing a reassignment if needed at the beginning is the simpler solution.

tombentley (approved):
Approved on the understanding that sufficient manual testing has been done. Do we have an issue for adding unit tests? IF not, please can we create one?

PR #31, opened 2021-04-08
Added more topic metadata logging

ppatierno:


internal/services/topic.go
```diff
@@ -108,6 +108,7 @@ func (ts *TopicService) Reconcile() (TopicReconcileResult, error) {
 	} else {
 		// canary topic already exists, check replicas assignments
 		log.Printf("The canary topic %s already exists\n", topicMetadata.Name)
+		logTopicMetadata(topicMetadata)
```
tombentley, line 111: Have we already logged the reassignment we're going to make in within the `if` block? If not perhaps we should, so we can see how we're changing them?

ppatierno: It's logged in the `alterAssignments` function

tombentley (approved):
LGTM

PR #32, opened 2021-04-08
Making topic internal using "__" prefix

### LINKED ISSUE #26: Making strimzi-canary topic as Kafka internal
The `strimzi-canary` topic should not be just a "normal" user topic so even visible through the admin client API by default.
It should be as an internal one in order to be filtered but visible if requested by the user.
In order to do that we should change the name using the `__` prefix so naming it `__strimzi-canary`.

ppatierno:
This PR fixes #26

### DIFF
--- a/internal/config/canary_config.go
+++ b/internal/config/canary_config.go
@@ -28,7 +28,7 @@ const (
 
 	// default values for environment variables
 	BootstrapServersDefault       = "localhost:9092"
-	TopicDefault                  = "__strimzi-canary"
+	TopicDefault                  = "__strimzi_canary"
 	ReconcileIntervalDefault      = 30000
 	ClientIDDefault               = "strimzi-canary-client"
 	ConsumerGroupIDDefault        = "strimzi-canary-group"
--- a/internal/config/canary_config.go
+++ b/internal/config/canary_config.go
@@ -28,7 +28,7 @@ const (
 
 	// default values for environment variables
 	BootstrapServersDefault       = "localhost:9092"
-	TopicDefault                  = "strimzi-canary"
+	TopicDefault                  = "__strimzi-canary"
 	ReconcileIntervalDefault      = 30000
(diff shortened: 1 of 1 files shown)


tombentley:
I don't have a problem with renaming this topic, but `__` doesn't do what you seem to think it does. Kafka itself doesn't really have a good notion of "internal" topics. The only true internal topics are `__consumer_offsets` and `__transaction_state`, but they're internal only because they're [hardcoded to be special](https://github.com/apache/kafka/blob/2.7.0/clients/src/main/java/org/apache/kafka/common/internals/Topic.java#L49), not because of the name. These are the only topics which won't be included by `Admin.listTopics` by default. All other topics are just topics as far as Kafka is concerned. From any _given point of view_ we can decide that certain topics are "internal" (e.g. Kafka Streams has its own notion of "internal" topics) but that internalness only really exists from that PoV. There's nothing stopping other clients interacting with them except for authorization. So `__` is really a figleaf which doesn't really hide anything.

If you really want to _prevent_ certain clients from seeing these topics then authz is the right way to do it.

scholzj:
@tombentley IIRC when I was digging into this because of the Topic Operator Kafka store, I think I would out that for example MM2 ignores the topics starting with `__` by default etc. So while the Admin API indeed will not mark it as internal, it might help in other cases to have it named like that.

ppatierno:
We know that `__` doesn't mark a topic as special in Kafka, I would like just follow a kind of pattern as the hardcoded "internal" topic __consumer_offsets and __transaction_state are using.

tombentley (approved):

tombentley:
@scholzj you're right it does do that by default. It's pretty adhoc though, since `__` doesn't officially mean anything, it's not even a documented "convention".

scholzj:
@tombentley Yeah, I know that it is no official thing. But in this case I guess it makes sense to not mirror it :-D

PR #34, opened 2021-04-12
Adding retrying mechanims to connect Kafka cluster on startup

### LINKED ISSUE #33: Canary panics when no Kafka cluster is available on startup
When the canary starts but there are no Kafka brokers available (as specified in the bootstrap servers), it fails with the following error due to a panic:

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x762be6]

goroutine 1 [running]:
github.com/Shopify/sarama.NewClusterAdminFromClient(0x0, 0x0, 0xc0001be270, 0xc0001be240, 0xc0001be210, 0xc000030420)
	/home/ppatiern/go/pkg/mod/github.com/!shopify/sarama@v1.27.2/admin.go:129 +0x26
github.com/strimzi/strimzi-canary/internal/services.NewTopicService(0xc000212000, 0x0, 0x0, 0x2)
	/home/ppatiern/github/strimzi-canary/internal/services/topic.go:39 +0x4d
main.main()
	/home/ppatiern/github/strimzi-canary/cmd/main.go:33 +0x16f
Process exiting with code: 2 signal: false
```

We should catch the proper error and using a retry mechanism until the Kafka cluster is available or giving up after a specified number of attempts.

ppatierno:
This PR fixes #33 adding a retrying mechanism to connect Kafka cluster during startup to avoid just exiting if the cluster is not available yet.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -57,7 +57,7 @@ func newClient(canaryConfig *config.CanaryConfig) sarama.Client {
 	config.Producer.Return.Successes = true
 	config.Producer.RequiredAcks = sarama.WaitForAll
 
-	backoff := services.NewBackoff(canaryConfig.BootstrapMaxAttempts, 5000)
+	backoff := services.NewBackoff(canaryConfig.BootstrapBackoffMaxAttempts, int64(canaryConfig.BootstrapBackoffScale))
 	for {
 		client, clientErr := sarama.NewClient([]string{canaryConfig.BootstrapServers}, config)
 		if clientErr == nil {
@@ -65,7 +65,7 @@ func newClient(canaryConfig *config.CanaryConfig) sarama.Client {
 		}
 		delay, backoffErr := backoff.Delay()
 		if backoffErr != nil {
-			log.Printf("Error connecting to the Kafka cluster after %d retries: %v", canaryConfig.BootstrapMaxAttempts, backoffErr)
+			log.Printf("Error connecting to the Kafka cluster after %d retries: %v", canaryConfig.BootstrapBackoffMaxAttempts, backoffErr)
 			os.Exit(1)
 		}
 		log.Printf("Error creating new Sarama client, retrying in %d ms: %v", delay, clientErr)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/Shopify/sarama"
 	"github.com/strimzi/strimzi-canary/internal/config"
@@ -55,7 +56,19 @@ func newClient(canaryConfig *config.CanaryConfig) sarama.Client {
 	config.Producer.Partitioner = sarama.NewManualPartitioner
 	config.Producer.Return.Successes = true
 	config.Producer.RequiredAcks = sarama.WaitForAll
-	// TODO: handling error
-	client, _ := sarama.NewClient([]string{canaryConfig.BootstrapServers}, config)
-	return client
+
+	backoff := services.NewBackoff(canaryConfig.BootstrapMaxAttempts, 5000)
+	for {
+		client, clientErr := sarama.NewClient([]string{canaryConfig.BootstrapServers}, config)
+		if clientErr == nil {
+			return client
+		}
+		delay, backoffErr := backoff.Delay()
+		if backoffErr != nil {
+			log.Printf("Error connecting to the Kafka cluster after %d retries: %v", canaryConfig.BootstrapMaxAttempts, backoffErr)
+			os.Exit(1)
+		}
+		log.Printf("Error creating new Sarama client, retrying in %d ms: %v", delay, clientErr)
+		time.Sleep(delay * time.Millisecond)
+	}
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -57,7 +57,7 @@ func newClient(canaryConfig *config.CanaryConfig) sarama.Client {
 	config.Producer.Return.Successes = true
 	config.Producer.RequiredAcks = sarama.WaitForAll
 
-	backoff := services.NewBackoff(canaryConfig.BootstrapBackoffMaxAttempts, int64(canaryConfig.BootstrapBackoffScale))
+	backoff := services.NewBackoff(canaryConfig.BootstrapBackoffMaxAttempts, canaryConfig.BootstrapBackoffScale*time.Millisecond, services.MaxDefault)
 	for {
 		client, clientErr := sarama.NewClient([]string{canaryConfig.BootstrapServers}, config)
 		if clientErr == nil {
@@ -68,7 +68,7 @@ func newClient(canaryConfig *config.CanaryConfig) sarama.Client {
 			log.Printf("Error connecting to the Kafka cluster after %d retries: %v", canaryConfig.BootstrapBackoffMaxAttempts, backoffErr)
 			os.Exit(1)
 		}
-		log.Printf("Error creating new Sarama client, retrying in %d ms: %v", delay, clientErr)
-		time.Sleep(delay * time.Millisecond)
+		log.Printf("Error creating new Sarama client, retrying in %d ms: %v", delay.Milliseconds(), clientErr)
+		time.Sleep(delay)
 	}
 }
--- a/internal/config/canary_config.go
+++ b/internal/config/canary_config.go
@@ -16,57 +16,61 @@ import (
 
 const (
 	// environment variables declaration
-	BootstrapServersEnvVar       = "KAFKA_BOOTSTRAP_SERVERS"
-	BootstrapMaxAttemptsEnvVar   = "KAFKA_BOOTSTRAP_MAX_ATTEMPTS"
-	TopicEnvVar                  = "TOPIC"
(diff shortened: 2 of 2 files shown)


ppatierno:
FYI @tomncooper @k-wall

cmd/main.go
```diff
@@ -55,7 +56,19 @@ func newClient(canaryConfig *config.CanaryConfig) sarama.Client {
...
+		if clientErr == nil {
+			return client
+		}
+		delay, backoffErr := backoff.Delay()
+		if backoffErr != nil {
+			log.Printf("Error connecting to the Kafka cluster after %d retries: %v", canaryConfig.BootstrapMaxAttempts, backoffErr)
+			os.Exit(1)
+		}
+		log.Printf("Error creating new Sarama client, retrying in %d ms: %v", delay, clientErr)
+		time.Sleep(delay * time.Millisecond)
```
tomncooper, line 72: isn't `backoff.Delay` already returning milliseconds?

tomncooper: Actually it is just a unitless `Duration`: `delay := time.Duration(b.scale * 1 << b.attempt)` so this makes sense.

cmd/main.go
```diff
@@ -55,7 +56,19 @@ func newClient(canaryConfig *config.CanaryConfig) sarama.Client {
 	config.Producer.Partitioner = sarama.NewManualPartitioner
 	config.Producer.Return.Successes = true
 	config.Producer.RequiredAcks = sarama.WaitForAll
-	// TODO: handling error
-	client, _ := sarama.NewClient([]string{canaryConfig.BootstrapServers}, config)
-	return client
+
+	backoff := services.NewBackoff(canaryConfig.BootstrapMaxAttempts, 5000)
```
tomncooper, line 60: Would it make sense to have the `scale` be configurable as well as the attempts? If you set attempts to a large number your backoff times could get huge?

cmd/main.go
ppatierno: yes but `sleep` takes a `Duration` and it's how you get a duration from the corresponding deley int as milliseconds.

tomncooper: `backoff.Delay()` returns a `time.Duration` not an int, but that doesn't have a unit so this line makes sense to make clear that these are in milliseconds.

ppatierno: see how the `time.Milliseconds` is defined

```go
const (
	Nanosecond  Duration = 1
	Microsecond          = 1000 * Nanosecond
	Millisecond          = 1000 * Microsecond
	Second               = 1000 * Millisecond
	Minute               = 60 * Second
	Hour                 = 60 * Minute
)
```

tomncooper: Good point.

cmd/main.go
ppatierno: Makes sense. Just did the change.

k-wall:
Doesn't this algorithm give unreasonably short backoffs for the the first few attempts, and then (unclamped as it it) excessively large values later?
Am I missing something?

=== RUN   TestBackoffMoreDelay2
attempt 0  Duration 0 ms 
attempt 1  Duration 0 ms 
attempt 2  Duration 0 ms 
attempt 3  Duration 0 ms 
attempt 4  Duration 0 ms 
attempt 5  Duration 0 ms 
attempt 6  Duration 0 ms 
attempt 7  Duration 0 ms 
attempt 8  Duration 1 ms 
attempt 9  Duration 2 ms 
attempt 10  Duration 5 ms 
attempt 11  Duration 10 ms 
attempt 12  Duration 20 ms 
attempt 13  Duration 40 ms 
attempt 14  Duration 81 ms 
attempt 15  Duration 163 ms 
attempt 16  Duration 327 ms 
attempt 17  Duration 655 ms 
attempt 18  Duration 1 seconds 
attempt 19  Duration 2 seconds 
attempt 20  Duration 5 seconds 
attempt 21  Duration 10 seconds 
attempt 22  Duration 20 seconds 
attempt 23  Duration 41 seconds 
attempt 24  Duration 1 minutes 23 seconds 
attempt 25  Duration 2 minutes 47 seconds 
attempt 26  Duration 5 minutes 35 seconds 
attempt 27  Duration 11 minutes 11 seconds 
attempt 28  Duration 22 minutes 22 seconds 
attempt 29  Duration 44 minutes 44 seconds 
--- PASS: TestBackoffMoreDelay2 (0.00s)

tomncooper (approved):

ppatierno:
@k-wall this is the same kind of backoff algorithm we use within Strimzi having 2 as the base of actually doubling the wait. what your test is doing? which scale did you set? Do you have a different idea about the backoff?

k-wall:
> @k-wall this is the same kind of backoff algorithm we use within Strimzi having 2 as the base of actually doubling the wait. what your test is doing? which scale did you set? Do you have a different idea about the backoff?
> Did you mean something that actually works in the other way around? Longer delay at the beginning and shorter later?
> If that's the case I can see your point. The Kafka cluster will take time to be up and running so we could start with long delay but then decreasing because more likely it becomes ready when time passes.
> I will explore that way then.

firstly, the code seems weird:

```
/ Delay computes a delay in milliseconds based on the current Backoff instance configuration
// Returns the delay in milliseconds and an error if the max attempts is reached, otherwise it's nil
func (b *Backoff) Delay() (time.Duration, error) {
```
Duration is defined as a number of nanoseconds.  The comment doesn't seem to respect this.

second, exponential backoff is fine, but it should be clamped to an upper value.  The point of a backoff algorithm is to balance the avoid a tight loop whist retrying in a timely fashion.  As soon as retry intervals get into hours, days, worse, it is pointless.  I'd clamp to say 5 minutes.

ppatierno:
> Duration is defined as a number of nanoseconds. The comment doesn't seem to respect this.

I agree on this, it should be a mistake as a newbie on Go when I wrote this code. I will fix this in a different PR.

> second, an increasing exponential backoff is fine, but it should be clamped to an upper value. The point of a backoff algorithm is to balance the avoid a tight loop whist retrying in a timely fashion. As soon as retry intervals get into hours, days, worse, it is pointless. I'd clamp to say 5 minutes.

I agree but not sure with what you are asking for.

Thinking more, isn't the idea of having an "inverse" backoff better?

k-wall:
> > Duration is defined as a number of nanoseconds. The comment doesn't seem to respect this.
> 
> I agree on this, it should be a mistake as a newbie on Go when I wrote this code. I will fix this in a different PR.

Thanks

> 
> > second, an increasing exponential backoff is fine, but it should be clamped to an upper value. The point of a backoff algorithm is to balance the avoid a tight loop whist retrying in a timely fashion. As soon as retry intervals get into hours, days, worse, it is pointless. I'd clamp to say 5 minutes.
> 
> I agree but not sure with what you are asking for.
> 
> Thinking more, isn't the idea of having an "inverse" backoff better?

Let's keep it simple - stick with a standard exponential backoff but add a clamp (upper bound), at 1 minute or 5 minutes.

tomncooper:
@k-wall Now that `scale` is configurable the max delay is implicit in the attempt/scale setting: `max-delay = scale * 2^max-attempt`

ppatierno:
Yes was going to say the same, we have an upper bound based on the max attempts, so why we should specify another parameter for it?

tomncooper:
You could add a sanity check that makes sure that equation is always less than some value? some fraction of a day?

ppatierno:
I opened #35 about backoff in correct Duration terms.

internal/config/canary_config.go
```diff
@@ -16,53 +16,61 @@ import (
...
+	BootstrapBackoffMaxAttemptsEnvVar = "KAFKA_BOOTSTRAP_BACKOFF_MAX_ATTEMPTS"
+	BootstrapBackoffScaleEnvVar       = "KAFKA_BOOTSTRAP_BACKOFF_SCALE"
+	TopicEnvVar                       = "TOPIC"
+	ReconcileIntervalEnvVar           = "RECONCILE_INTERVAL_MS"
+	ClientIDEnvVar                    = "CLIENT_ID"
+	ConsumerGroupIDEnvVar             = "CONSUMER_GROUP_ID"
+	TLSEnabledEnvVar                  = "TLS_ENABLED"
+	ProducerLatencyBucketsEnvVar      = "PRODUCER_LATENCY_BUCKETS"
+	EndToEndLatencyBucketsEnvVar      = "ENDTOEND_LATENCY_BUCKETS"
+	ExpectedClusterSizeEnvVar         = "EXPECTED_CLUSTER_SIZE"
```
tombentley, line 29: Can we avoid aligning these? I know it looks a bit prettier, but as in this case it makes for much noisier diffs.

tombentley (approved):

internal/config/canary_config.go
ppatierno: @tombentley I would be with you but it's Go stuff and it's good practice in the Go land. Each IDE align in this way automatically because of this.

ppatierno:
@k-wall can you have another pass please now that #35 is merged and this branch rebased on that? Thanks!

k-wall (approved):

k-wall:
Looks good to me.

PR #35, opened 2021-04-12
Fixed the backoff with the nanoseconds Duration meaning

ppatierno:
The current backoff stuff relying on a misleading meaning of the delay in milliseconds.
This PR reverts everything into Duration terms which defines times in nanoseconds as expected by sleep functions as well.

### DIFF
--- a/internal/services/backoff.go
+++ b/internal/services/backoff.go
@@ -12,14 +12,16 @@ import (
 )
 
 const (
-	ScaleDefault       = 200
+	ScaleDefault       = 200 * time.Millisecond
+	MaxDefault         = 5 * time.Minute
 	MaxAttemptsDefault = 6
 )
 
 // Backoff encapsulates computing delays for an exponential back-off, when an operation has to be retried
 type Backoff struct {
 	maxAttempts int
-	scale       int64
+	scale       time.Duration
+	max         time.Duration
 	attempt     int
 }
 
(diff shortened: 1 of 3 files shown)


ppatierno:
FYI @k-wall @tomncooper

internal/services/backoff.go
```diff
@@ -47,6 +58,9 @@ func (b *Backoff) Delay() (time.Duration, error) {
 		return 0, &MaxAttemptsExceeded{}
 	}
 	delay := time.Duration(b.scale * 1 << b.attempt)
+	if delay > b.max {
```
k-wall, line 61: what about overflow from the line above?

tombentley (approved):

k-wall:
Looks much more sane :)

tomncooper (approved):

internal/services/backoff.go
k-wall: if delay < 0 panic(...) would be enough for me. The ensuing kube restart would put the code back onto the right track.

ppatierno: @k-wall good point, I changed the code a little bit for checking overflow.

internal/services/backoff.go
```diff
@@ -57,9 +58,16 @@ func (b *Backoff) Delay() (time.Duration, error) {
 	if b.attempt == b.maxAttempts {
 		return 0, &MaxAttemptsExceeded{}
 	}
-	delay := time.Duration(b.scale * 1 << b.attempt)
-	if delay > b.max {
+	var delay time.Duration
+	// check that it doesn't overflow on int64
+	delayFloat := float64(b.scale) * math.Pow(2, float64(b.attempt))
```
k-wall, line 63: haven't you just swapped the possibility of an int64 overflow for an float64 one? :)

k-wall: ignore that, I'm wrong

ppatierno: That's true but I was trying to be a little bit realistic :-)
Anyway let's back just to panic then.

k-wall: Approved, but to my taste, I prefer `if delay < 0 return 0, &MaxAttemptsExceeded{}" // guard overflow` rather than introduce the floating point complication.

k-wall (approved):

internal/services/backoff.go
ppatierno: Happy about that. Going to change.

ppatierno: @k-wall I finally simplified it but returning a more specific `BackoffDelayOverflow` error.

k-wall (approved):

PR #37, opened 2021-04-13
Added backoff retry mechanism on expected cluster size not met

### LINKED ISSUE #36: Canary panics and crash looping when cluster size is not the expected yet
When the expected cluster size is enabled, the canary just panics and exits if the expected number of brokers is not what is taken from the cluster description.
The canary should wait and retry instead of crash looping even because this action is doing at startup.

ppatierno:
This PR fixes #36

### DIFF
--- a/internal/services/topic.go
+++ b/internal/services/topic.go
@@ -7,6 +7,7 @@
 package services
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"strconv"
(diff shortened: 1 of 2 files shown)


internal/services/topic.go
```diff
@@ -34,6 +35,13 @@ type TopicService struct {
 	initialized  bool
 }
 
+// ErrExpectedClusterSize defines the error raised when the expected cluster size is not met
+type ErrExpectedClusterSize struct{}
+
+func (e *ErrExpectedClusterSize) Error() string {
+	return fmt.Sprintf("Current cluster size doesn't met the expected one")
```
tombentley, line 42:
```suggestion
	return fmt.Sprintf("Current cluster size differs from the expected size")
```

Should we include the values too?

tombentley (approved):

PR #40, opened 2021-04-19
Added Dockerfile and Makefile for compiling and building image

ppatierno:
This PR is a first starting point of having Dockerfile and Makefile to build the canary image.
It replaces #15

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,24 @@
+IMAGE_REGISTRY ?= quay.io
+IMAGE_ORG ?= strimzi
+IMAGE_REPO ?= strimzi-canary
+IMAGE_TAG ?= 0.0.1
+
+BINARY ?= strimzi-canary
+
+go_build:
+	echo "Building Golang binary ..."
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,24 @@
+IMAGE_REGISTRY ?= quay.io
+IMAGE_ORG ?= strimzi
+IMAGE_REPO ?= strimzi-canary
```
scholzj, line 3: Should it be just `canary`?

Makefile
```diff
@@ -0,0 +1,24 @@
...
+IMAGE_TAG ?= 0.0.1
```
scholzj, line 4: I think we should follow the names from other repositories:

```
DOCKER_REGISTRY    ?= docker.io
DOCKER_ORG         ?= $(USER)
DOCKER_TAG         ?= latest
```

Possibly with the same default values.

ppatierno:
@scholzj I addressed your comments, maybe you can have another pass?

scholzj (approved):

PR #54, opened 2021-05-18
Added TLS connection support

### LINKED ISSUE #9: Add TLS support
Add TLS support to the canary. Currently the canary config contains an environment variable to enable TLS. However it is disabled by default and not implemented even if enabled.

ppatierno:
This PR fixes #9 
It allows to specify the TLS CA certificate, client and key directly in the corresponding env vars or using them to specify a path that could be mounted as a Secret (as it could be done by the Strimzi operator in the future and as it's already done for other components).
The code just configures the Sarama library with the corresponding TLS configuration.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -91,9 +91,6 @@ func newClient(canaryConfig *config.CanaryConfig) (sarama.Client, error) {
 	}
 
 	if canaryConfig.TLSEnabled {
-		if canaryConfig.TLSCACert == "" {
-			glog.Fatalf("TLS enabled but the CA certificate is not provided")
-		}
 		config.Net.TLS.Enable = true
 		if config.Net.TLS.Config, err = security.NewTLSConfig(canaryConfig); err != nil {
 			glog.Fatalf("Error configuring TLS: %v", err)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -92,7 +92,7 @@ func newClient(canaryConfig *config.CanaryConfig) (sarama.Client, error) {
 
 	if canaryConfig.TLSEnabled {
 		if canaryConfig.TLSCACert == "" {
-			glog.Fatalf("TLS enabled but at leats the CA certificate is not provided")
+			glog.Fatalf("TLS enabled but the CA certificate is not provided")
 		}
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -89,6 +90,16 @@ func newClient(canaryConfig *config.CanaryConfig) (sarama.Client, error) {
 		sarama.Logger = log.New(os.Stdout, "[Sarama] ", log.LstdFlags)
 	}
 
+	if canaryConfig.TLSEnabled {
+		if canaryConfig.TLSCACert == "" {
+			glog.Fatalf("TLS enabled but at leats the CA certificate is not provided")
```
tombentley, line 95:
```suggestion
			glog.Fatalf("TLS enabled but the CA certificate is not provided")
```

tombentley (approved):

internal/security/security.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/golang/glog"
+	"github.com/strimzi/strimzi-canary/internal/config"
+)
+
+func NewTLSConfig(canaryConfig *config.CanaryConfig) (*tls.Config, error) {
+	tlsConfig := &tls.Config{}
+
+	if canaryConfig.TLSCACert != "" {
+		if caCert, err := loadCertKey(config.TLSCACertEnvVar, canaryConfig.TLSCACert); err == nil {
+			tlsConfig.RootCAs = x509.NewCertPool()
```
k-wall, line 24: I think it is more flexible to merge the provided trust with the system store.

```
tlsConfig.RootCAs, _ := x509.SystemCertPool()
if tlsConfig.RootCAs == nil {
	tlsConfig.RootCAs = x509.NewCertPool()
}
```

k-wall (commented):
Looks good on the whole, just a few comments left.

ppatierno:
@k-wall I see your points, I will double check but unless we change the base image for the canary we are not going to have a system default certs pool because we are using `FROM scratch` in the Dockerfile. Anyway, it's a good point because the canary binary could be used on its own locally or users could want to use their own Unix image as base and not "scratch".

k-wall (approved):
looks reasonable to me.

PR #61, opened 2021-05-26
Sarama using logging in UTC

ppatierno:
This PR enables Sarama to use UTC time on logging.


tombentley:
IIRC some Strimzi users set `TZ` env var to customise the timezone of the logs. Can we do the same here?

ppatierno:
Good point. Using just `TZ` is more general otherwise people could set TZ=CEST having canary logging in CEST but then Sarama in UTC. I will close this PR, setting the TZ is just part of the deployment.

PR #62, opened 2021-05-28
Add a simple grafana-dashboard

elakito:
As requested during the discussion on the slack channel, I am creating this PR to show how it looks. There is a simple instruction to try it out and a screenshot. For now, all files are placed in tmp folder.


elakito:
@ppatierno it has been a while and I don't know if the dashboard needs to be adjusted. If some adjustment is required, I can look at it. I think it would be nice to see what Canary is doing. So, if you could review it, that would be great.
thanks, aki

tmp/canary-monitor.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: canary-metrics
+  labels:
+    app: strimzi
+spec:
+  selector:
+    matchLabels:
+      strimzi.io/kind: Canary
```
ppatierno, line 10: I am not sure we should use the `strimzi.io/kind` as label for selecting the canary service. We use that label in the Strimzi world related to existing "kind" so corresponding custom resources (i.e. Kafka, KafkaConnect and so on). Maybe just using the `app: canary` label? @scholzj what's your opinion on this?

tmp/canary-monitor.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespaceSelector:
+    matchNames:
+      - my-kafka
```
ppatierno, line 13: if this is an example you should be clear in the description that the namespace should be `my-kafka` otherwise metrics discovery will not work.

tmp/canary.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: canary-service
+  labels:
+    app: canary
+    strimzi.io/kind: Canary
```
ppatierno, line 7: see previous comment about this label.

tmp/grafana-dashboards/strimzi-canary.json
```diff
@@ -0,0 +1,1012 @@
...
+      "sparkline": {
+        "fillColor": "rgba(31, 118, 189, 0.18)",
+        "full": false,
+        "lineColor": "rgb(31, 120, 193)",
+        "show": false
+      },
+      "tableColumn": "",
+      "targets": [
+        {
+          "expr": "sum(strimzi_canary_records_produced_total)",
```
ppatierno, line 167: should it be `_consumed_` ?

tmp/grafana-dashboards/strimzi-canary.json
```diff
@@ -0,0 +1,1012 @@
...
+      "sparkline": {
+        "fillColor": "rgba(31, 118, 189, 0.18)",
+        "full": false,
+        "lineColor": "rgb(31, 120, 193)",
+        "show": false
+      },
+      "tableColumn": "",
+      "targets": [
+        {
+          "expr": "sum(rate(strimzi_canary_records_consumed_total[5m])*60)",
```
ppatierno, line 289: why reporting per mins and not per secs?

tmp/canary-monitor.yaml
elakito: Yes. I should add a comment. I thought strimzi quick-guide was also using namespace my-kafka. But it is actually using `my-kafka-project`. So, I can change it to `my-kafka-project` to align it to the quick guide add also a comment above to adjust it for the target clusters. How do you think?

tmp/grafana-dashboards/strimzi-canary.json
elakito: For the default interval value (30-sec) used in canary, using rec/sec lead to an ugly fraction number, so I thought using rec/min would be more appropriate.

tmp/grafana-dashboards/strimzi-canary.json
elakito: yes. oh. It looks like this one and for the average recording producing/consuming latency panels, the name is swapped. I'll fix them.

tmp/canary-monitor.yaml
Frawless: Agree, we should use label which we have in `020-Deployment.yaml`

tmp/canary-monitor.yaml
```diff
@@ -0,0 +1,16 @@
...
+  podMetricsEndpoints:
+  - path: /metrics
+    port: rest-api
```
Frawless, line 16:
```suggestion
    port: metrics
```

tmp/canary.yaml
```diff
@@ -0,0 +1,39 @@
...
+      labels:
+        app: canary
+        strimzi.io/kind: Canary
+    spec:
+      containers:
+      - name: canary
+        image: strimzi-canary:0.0.1
+        ports:
+        - containerPort: 8080
+          name: rest-api
```
Frawless, line 35: I think we can only add this part to `packaging/install` because deployment and service is already defined there

tmp/grafana-dashboards/strimzi-canary.json
```diff
@@ -0,0 +1,1012 @@
...
+    {
+      "cacheTimeout": null,
+      "colorBackground": false,
+      "colorValue": false,
+      "colors": [
+        "#299c46",
+        "rgba(237, 129, 40, 0.89)",
+        "#d44a3a"
+      ],
+      "datasource": "Prometheus",
```
Frawless, line 78:
```suggestion
      "datasource": "${DS_PROMETHEUS}",
```

we use this format in Strimzi examples so I think we should follow the same pattern

tmp/grafana-dashboards/strimzi-canary.json
```diff
@@ -0,0 +1,1012 @@
...
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "7.3.7"
```
Frawless, line 17: this is the minimal version needed?

ppatierno:
@elakito do you think you can address comments from Jakub. It would be great to finally merge this one now that we had a first canary release :-)

elakito:
@ppatierno Ok. I'll have a look and go through with the comments...

tmp/grafana-dashboards/strimzi-canary.json
elakito: I took these properties from the dashboards of strimzi-kafka-operator https://github.com/strimzi/strimzi-kafka-operator/blob/main/examples/metrics/grafana-dashboards/strimzi-kafka.json#L12

tmp/grafana-dashboards/strimzi-canary.json
elakito: ok

tmp/canary-monitor.yaml
elakito: ok.

tmp/canary.yaml
elakito: ok.

tmp/canary-monitor.yaml
elakito: ok.

tmp/canary.yaml
elakito: yes. When I opened this PR some months ago, the installation folder was not there, so I created this temporary deployment descriptor in this PR. I'll update it to the current state.

elakito:
Hi all,
I adjusted the pod monitor yaml canary-monitor.yaml to reflect the comments and it works with strimzi-canary's current package/install/*.yaml files. This adjustment also deleted the tmp/canary.yaml and updated tmp/README.txt.

To finalize this PR, could you suggest where these files currently should be placed?

tmp
├── canary-monitor.yaml
├── grafana-dashboard.png
├── grafana-dashboards
    └── strimzi-canary.json

Maybe create a top level folder "metrics" and place these files there?

ppatierno:
@elakito first of all thank you very much for coming back to this!
About the files, I would like to follow the same pattern we have with something similar in the strimzi-kafka-operator repo.
We should have a `packaging/examples/metrics` folder with a `grafana-dashboards` subfolder where putting the `strimzi-canary.json` and maybe a `prometheus-install` subfolder containing the `canary-monitor.yaml`.
About the `tmp/README.txt` I would create a section about Monitoring in the main README file.
Not sure we really need a screenshot of the dashboard, we don't have something like this in the operator.
@scholzj what do you think about all the above?

tmp/README.txt
```diff
@@ -0,0 +1,21 @@
+Some resources for experimenting with strimzi-canary
+
+1. Preparation
+Make sure that you have your strimzi-canary docker image. If not, build strimzi-canary and
+make its image available to your cluster.
```
ppatierno, line 5: not sure we really need this pre-requisite. User can use the official canary image, not building a custom one, or?

scholzj:
I don't think I have a strong preference. I guess @ppatierno suggestion looks good and mirrors what we have in operators. I can also imagine the README just being in the directory with the dashboard and PodMonitor. But not strong opinions.

scholzj (approved):

scholzj:
Thanks for the work.

@Frawless Could you please look as well since you commented before on this PR.

Frawless (approved):

PR #64, opened 2021-07-02
Static container

henryZrncik:
Github action for pull and push to the main branch. Some basic test to confirm correct behaviour of canary topic and endpoints. 
Tests use static docker container described in compose instead of test containers due to problematic behaviour of Canary if other than default ports for Kafka and Zookeeper (9092, 2181). I can describe it in more details if needed.


test/base_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"context"
+	"log"
+	"net/http"
+	"testing"
+	"time"
+)
+
+/* TODO: current version doesn't use dynamic port allocation due to Canary not running properly on Kafka mapped to any other port than default  9092
++ what is later propagated as series of unwanted behaviours, for example not having open port on 9092 ends in failed attempt to connect to this port despite other port being set in configuration
+*/
```
ppatierno, line 15: I don't see this problem. I have just run a local Kafka cluster with one broker listening on port 9093 (so not 9092) and configuring the canary with the KAFKA_BOOTSTRAP_SERVER env var at localhost:9093 and it just connects and works.

test/compose-kafka-zookeeper.yaml
```diff
@@ -0,0 +1,18 @@
+version: "3"
+services:
+  zookeeper:
+    image: wurstmeister/zookeeper:latest
```
ppatierno, line 4: would be better using strimzi images for this. Does them work fine right @scholzj? I remember that you suggested it to the Quarkus team for their tutorials.

scholzj: Yeah ... I would say so: https://github.com/scholzj/strimzi-compose-up

test/base_test.go
henryZrncik: in my case (kafka running on port 9093, env variable set:  KAFKA_BOOTSTRAP_SERVERS=localhost:9093, and port 9092 closed ) it ends up like this: 
```
E0707 13:42:01.424011   69927 topic.go:104] Error describing cluster: dial tcp 127.0.0.1:9092: connect: connection refused
F0707 13:42:01.424028   69927 canary_manager.go:76] Error starting manager: dial tcp 127.0.0.1:9092: connect: connection refused
```

and in same situation as above but with kafka running on both 9092 and 9093 problem appears only after 2 and a half minute:
```
Error joining the consumer group: Maximum number of attempts exceeded
exit status 255
```

ppatierno: when the canary starts, it prints the configuration it's getting from the env vars. Checks what's your configuration and if it's not what you expect, you have some problem passing env vars but it can't be canary related.

henryZrncik: configuration is according to expectation, it is printed correctly.  (For upper example )
```
Starting Strimzi canary tool with config: {BootstrapServers:localhost:9093, ...
```

ppatierno: can you provide a cleaner explanation and steps to reproduce what you are seeing please?

test/base_test.go
```diff
@@ -0,0 +1,125 @@
+package test
```
fgiorgetti, line 1: One hint, which has been very useful to us, is to define build tags.
Its handy for running just a specific subset of your test suites.

test/base_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"context"
+	"github.com/segmentio/kafka-go"
```
ppatierno, line 5: we should use just the Sarama library and not adding one more dependency for another Kafka client.

test/base_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+)
+
+
+const (
+	httpUrlPrefix                   = "http://localhost:8080"
+	metricsEndpoint                 = "/metrics"
+	canaryTopicName                 = "__strimzi_canary"
+	metricServerUpdateTimeInSeconds = 30
+	kafkaServer = "localhost:9092"
```
ppatierno, line 19: align

fgiorgetti, line 19: Quick question. This local HTTP and the kafka server will always run at
the same host where the test suite is executed from? Or are you anticipating
scenarios where test suite runs in a given machine but kafka is running 
somewhere else?

test/container_manager/container_manager.go
```diff
@@ -0,0 +1,62 @@
...
+
+func CreateManager() ContainerManager {
+	var cManipulation ContainerManager = &controller{}
+	return cManipulation
+}
+
+
+func (c *controller) StartCanary() {
+	log.Println("starting canary")
+	go createCanary()
```
ppatierno, line 59: if we are doing an end-to-end test, as we start ZK and K as containers shouldn't be do kind of same thing with canary? Maybe not container but local application but not re-writing the canary main logic for starting all the services (you already mentioned the DRY principle).

test/container_manager/canary_utils.go
```diff
@@ -0,0 +1,109 @@
...
+// TODO goes against DRY principle, possible refactor of cmd/main to put more of code into none main module
+func createCanary() {
+	canaryConfig := config.NewCanaryConfig()
+
+	// Always log to stderr by default
+	if err := flag.Set("logtostderr", "true"); err != nil {
+		glog.Errorf("Error on setting logtostderr to true")
+	}
+	flag.Set("v", strconv.Itoa(canaryConfig.VerbosityLogLevel))
+	flag.Parse()
```
fgiorgetti, line 39: Please evaluate if moving flag parsing to a `TestMain(*testing.M)` helps.
The benefit is that it happens at package level, and so you avoid calling
`flag.Parse()` different tests within the same package.

test/container_manager/canary_utils.go
```diff
@@ -0,0 +1,109 @@
...
+	producerService := services.NewProducerService(canaryConfig, client)
+	consumerService := services.NewConsumerService(canaryConfig, client)
+
+	canaryManager := workers.NewCanaryManager(canaryConfig, topicService, producerService, consumerService)
+	canaryManager.Start()
+
+	sig := <-signals
+	glog.Infof("Got signal: %v", sig)
+	canaryManager.Stop()
+	httpServer.Stop()
```
fgiorgetti, line 64: Guess it might be useful to defer these two calls to `Stop()` right
after the servers have been started.

test/base_test.go
fgiorgetti: Another suggestion for you guys to evaluate, would be moving
those tests from `/test` to some more specific directory, like
`/test/base` or `/test/e2e/base`.

fgiorgetti (commented):
@henryZrncik First of all, great work!
@see-quick I have provided a few comments and suggestions for you guys to evaluate.
Not sure if they are all applicable or relevant, but might be interesting things to discuss at least.

test/container_manager/container_manager.go
Frawless: Would be possible to do something similar as we have in the bridge tests? Run canary locally or somewhere and work with it?

test/base_test.go
henryZrncik: Yes at current situation I expect they run at the same host, but in future I can imagine adding scenario with them separated even more than just by having kafka running inside container.

test/container_manager/canary_utils.go
henryZrncik: both this and next one recommendation are sort of resolved, as I refactored this part of code (replace by running application as different process, which was another suggestion by Paolo)

test/container_manager/container_manager.go
henryZrncik: Something like that is my original solution, but methods which are used to start canary are in the main package, so in Go I have no other choice than Copy Paste, so instead of it I start Canary separately via go cmd. line (Exec).

test/base/base_runner_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+// global variables for accessing container information
+var (
+	controller service_manager.ServiceManager
+)
+
+func TestMain(m *testing.M) {
+
+
+
```
ppatierno, line 18: do we need all these spaces? :-)

test/service_manager/service_manager.go
```diff
@@ -0,0 +1,71 @@
...
+
+func CreateManager() ServiceManager {
+	var cManipulation ServiceManager = &controller{}
+	return cManipulation
+}
+
+
+func (c *controller) StartCanary() {
+	log.Println("starting canary")
+	// Output is not necessary but
```
ppatierno, line 66: but ... ? missing info here?

ppatierno (changes requested):

go.mod
```diff
@@ -4,6 +4,9 @@ go 1.13
 
 require (
 	github.com/Shopify/sarama v1.29.0
+	github.com/docker/go-connections v0.4.0
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
 	github.com/prometheus/client_golang v1.8.0
+	github.com/segmentio/kafka-go v0.4.16
```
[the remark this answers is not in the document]
henryZrncik: Sorry I forgot to remove it, will be fixed in a while with rest of items.

go.sum
```diff
@@ -1,7 +1,13 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78 h1:w+iIsaOQNcT7OZ575w+acHgRric5iCyQh+xv+KJ4HB8=
+github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78/go.mod h1:LmzpDX56iTiv29bbRTIsUNlaFfuhWRQBWjQdVyAevI8=
```
ppatierno, line 4: what's adding this Azure stuff?

go.sum
```diff
@@ -1,7 +1,13 @@
...
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/Knetic/govaluate v3.0.1-0.20171022003610-9aa49832a739+incompatible/go.mod h1:r7JcOSlj0wfOMncg0iLm8Leh48TZaKVeNIfJntJ2wa0=
+github.com/Microsoft/go-winio v0.4.11 h1:zoIOcVf0xPN1tnMVbTtEdI+P8OofVk3NObnwOQ6nK2Q=
+github.com/Microsoft/go-winio v0.4.11/go.mod h1:VhR8bwka0BXejwEJY73c50VrPtXAaKcyvVC4A4RozmA=
+github.com/Microsoft/hcsshim v0.8.6 h1:ZfF0+zZeYdzMIVMZHKtDKJvLHj76XCuVae/jNkjj0IA=
+github.com/Microsoft/hcsshim v0.8.6/go.mod h1:Op3hHsoHPAvb6lceZHDtd9OkTew38wNoXnJs8iY7rUg=
```
ppatierno, line 10: same here, what's adding this MS stuff?

go.sum
```diff
@@ -66,19 +85,31 @@ github.com/franela/goblin v0.0.0-20200105215937-c9ffbefa60db/go.mod h1:7dvUGVsVB
...
 github.com/go-kit/kit v0.10.0/go.mod h1:xUsJbQ/Fp4kEt7AFgCuvyX4a71u8h9jB8tj/ORgOZ7o=
 github.com/go-logfmt/logfmt v0.3.0/go.mod h1:Qt1PoO58o5twSAckw1HlFXLmHsOX5/0LbT9GBnD5lWE=
 github.com/go-logfmt/logfmt v0.4.0/go.mod h1:3RMwSq7FuexP4Kalkev3ejPJsZTpXXBr9+V4qmtdjCk=
 github.com/go-logfmt/logfmt v0.5.0/go.mod h1:wCYkCAKZfumFQihp8CzCvQ3paCTfi41vtzG1KdI/P7A=
+github.com/go-playground/assert/v2 v2.0.1/go.mod h1:VDjEfimB/XKnb+ZQfWdccd7VUvScMdVu0Titje2rxJ4=
+github.com/go-playground/locales v0.13.0/go.mod h1:taPMhCMXrRLJO55olJkUXHZBHCxTMfnGwq/HNwmWNS8=
+github.com/go-playground/universal-translator v0.17.0/go.mod h1:UkSxE5sNxxRwHyU+Scu5vgOQjsIJAF8j9muTVoKLVtA=
+github.com/go-playground/validator/v10 v10.2.0/go.mod h1:uOYAAleCW8F/7oMFd6aG0GOhaH6EGOAJShg8Id5JGkI=
+github.com/go-redis/redis v6.15.9+incompatible h1:K0pv1D7EQUjfyoMql+r/jZqCLizCGKFlFgcHWWmHQjg=
+github.com/go-redis/redis v6.15.9+incompatible/go.mod h1:NAIEuMOZ/fxfXJIrKDQDz8wamY7mA7PouImQ2Jvg6kA=
```
ppatierno, line 104: redis? from where it's coming?

ppatierno (changes requested):
I have concerns about folders and package naming. 
Why we are using a base folder under test for example? 
Why we are calling this integration e2e tests using a file named base_test.go? Isn't it integration_test.go or e2e_test.go much better? At some point we could have unit tests as well.

test/base/base_test.go
```diff
@@ -0,0 +1,131 @@
...
+		wantResponseStatus := testInput.expectedStatusCode
+		gotResponseStatus := resp.StatusCode
+		if wantResponseStatus != gotResponseStatus {
+			t.Errorf("endpoint: %s expected response code: %d obtained: %d" ,completeUrl,wantResponseStatus,gotResponseStatus  )
+		}
+		log.Printf("endpoint:  %s, responded with expected status code %d\n", testInput.endpoint, testInput.expectedStatusCode)
+	}
+}
+
+func TestMetricServerContentUpdating(t *testing.T) {
```
ppatierno, line 104: I see this test checking for some Prometheus related metrics, why not canary related metrics?

ppatierno (changes requested):

test/base/base_test.go
```diff
@@ -0,0 +1,131 @@
...
+			t.Error(err.Error())
+		}
+
+
+		if !test.IsTopicPresent(canaryTopicName, topics) {
+			t.Errorf("%s is not present", canaryTopicName)
+		}
+
+		// consume single message
+		consumer, _ := sarama.NewConsumer(brokers, nil)
```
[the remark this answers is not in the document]
henryZrncik: yes I can it was just that these were originally 2 different tests, thus they had their own consumer. Now it is just 1 consumer used for both of them.

test/base/base_test.go
henryZrncik: I firstly needed to resolve occasional problem with running Canary, Now solved so this test is added as well.

test/base/base_test.go
```diff
@@ -0,0 +1,131 @@
...
+		partitionConsumer, _ := consumer.ConsumePartition(canaryTopicName, 0, 0)
```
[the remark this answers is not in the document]
henryZrncik: changed as you suggested.

go.sum
henryZrncik: All of These problems are due to using TestContainers in test, for now I replaced their usage with simple docker-compose and exec.

.github/workflows/acceptancetest.yml
```diff
@@ -2,7 +2,7 @@ name: acceptancetest
 
 on:
   push:
-    branches: [main]
+    branches: [main, ff]
```
ppatierno: what's this `ff` branch?

test/e2e_runner_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+var (
+	serviceManager *service_manager.ServiceManager
+)
+
+// returning into the root from /test
+func init() {
+	_, filename, _, _ := runtime.Caller(0)
+	dir := path.Join(path.Dir(filename), "..")
+	err := os.Chdir(dir)
```
ppatierno, line 20: can you explain why this code is needed as init?

test/e2e_test.go
```diff
@@ -0,0 +1,166 @@
...
+const (
+	httpUrlPrefix   = "http://localhost:8080"
+	metricsEndpoint = "/metrics"
+)
+
+/* test checks for following:
+*  the presence of canary topic,
+*  liveliness of topic (messages being produced),
+*/
+func TestCanaryTopicLiveliness(t *testing.T) {
```
ppatierno, line 26: I guess you meant `Liveness` here? In the comment as well?

test/e2e_test.go
```diff
@@ -0,0 +1,166 @@
...
+	// test itself.
+	go func() {
+		config := sarama.NewConfig()
+		config.Consumer.Return.Errors = true
+		config.Consumer.Offsets.Initial =  sarama.OffsetOldest
+		ctx := context.Background()
+
+		//kafka end point
+		brokers := []string{serviceManager.KafkaBrokerAddress}
+		//get broker
```
ppatierno, line 44: get broker?

test/e2e_test.go
```diff
@@ -0,0 +1,166 @@
...
+*  liveliness of topic (messages being produced),
+*/
+func TestCanaryTopicLiveliness(t *testing.T) {
+	log.Println("TestCanaryTopic test starts")
+
+	// setting up timeout and handler for consumer group, there is no need to wait much longer then is the retention time.
+	timeout := time.After(time.Second * 10)
+	handler := exampleConsumerGroupHandler{}
+	handler.mutexWritePartitionPresence = &sync.Mutex{}
+	handler.consumingDone = make(chan bool)
```
ppatierno, line 33: cannot do this initialization inside the above declaration ?

test/e2e_test.go
```diff
@@ -0,0 +1,166 @@
...
+		group, err := sarama.NewConsumerGroup([]string{serviceManager.KafkaBrokerAddress}, "faq-g9", config)
+		if err != nil {
+			panic(err)
+		}
+
+		// set up client for getting partition count on canary topic
+		client, _ := sarama.NewClient([]string{serviceManager.KafkaBrokerAddress},config )
+		listOfPartitions, _ := client.Partitions(serviceManager.TopicTestName)
+		var partitionsCount =  len(listOfPartitions)
+		handler.partitionsConsumptionSlice = make([]bool, partitionsCount)
```
ppatierno, line 68: I think that for getting topic existence and partitions you should use the same approach as the topic.go file. It's using the admin client to describe topics, getting just the one you need to check, without then looping in a collection and getting metadata with partitions info as well.

test/service_manager/service_manager.go
```diff
@@ -0,0 +1,166 @@
...
+	manager.KafkaBrokerAddress = kafkaBrokerAddress
+	return manager
+}
+
+func (c *ServiceManager) StartCanary() {
+	log.Println("Starting Canary")
+	c.setUpCanaryParamsViaEnv()
+	myCmd := exec.Command("go", "run",  c.pathToCanaryMain )
+	//myCmd.Stdout = os.Stdout
+	//myCmd.Stderr = os.Stderr
```
ppatierno, line 76: remove commented lines

test/service_manager/service_manager.go
```diff
@@ -0,0 +1,166 @@
...
+	log.Println("start waiting for broker")
+	timeout := time.After(10 * time.Second)
+	brokerIsReadyChannel := make(chan bool)
+
+	go func() {
+		configuration := sarama.NewConfig()
+		brokers := []string{c.KafkaBrokerAddress}
+
+		for ;; {
+			consumer, err := sarama.NewConsumer(brokers, configuration)
```
ppatierno, line 97: you are not consuming here, just use a `NewClient` instead of consumer, or `NewAdminClient`.
There is something similar in the main.go with a backoff mechanism that could be useful as well.

test/service_manager/service_manager.go
```diff
@@ -0,0 +1,166 @@
...
+	log.Println("start waiting for canary setup")
+	timeout := time.After(10 * time.Second)
+	canaryIsReadyChannel := make(chan bool)
+
+	go func() {
+		configuration := sarama.NewConfig()
+		brokers := []string{c.KafkaBrokerAddress}
+
+		for ;;{
+			consumer, err := sarama.NewConsumer(brokers, configuration)
```
ppatierno, line 130: same as already mentioned before use the admin client to do the check on the canary topic and its existence.

ppatierno (changes requested):

.github/workflows/acceptancetest.yml
henryZrncik: apologise, i used it verify github actions without pushing on this branch, it will be removed in a while.

test/e2e_test.go
henryZrncik: I have created constructor for handler, and moved timeout to selector, but besides that timeout must stay within method.

test/e2e_runner_test.go
henryZrncik: Move one step upper (to root of project) in path from /test. But I can simply count with that.

test/e2e_test.go
```diff
@@ -0,0 +1,155 @@
...
+		// consume single message
+		group, err := sarama.NewConsumerGroup([]string{serviceManager.KafkaBrokerAddress}, "faq-g9", config)
+		if err != nil {
+			panic(err)
+		}
+
+		// set up client for getting partition count on canary topic
+		client, _ := sarama.NewClient([]string{serviceManager.KafkaBrokerAddress},config )
+		listOfPartitions, _ := client.Partitions(serviceManager.TopicTestName)
+		var partitionsCount =  len(listOfPartitions)
```
ppatierno, line 56: you can get the partition count from the `topicsDescriptionList[0].Partitions`, no need for a new client just for that.

test/service_manager/service_manager.go
```diff
@@ -0,0 +1,156 @@
...
+
+// Implementation of Service Manager
+type ServiceManager struct {
+	CanaryConfig
+	Paths
+}
+
+// Configurations of canary that is manipulated in e2e tests
+type CanaryConfig struct {
+	RetentionTime      string
```
ppatierno, line 20: Retention? Isn't it the Reconcile time? As I see it's set on canary? This should be changed in the service manager as well.

test/test_utilities.go
```diff
@@ -0,0 +1,86 @@
...
+func parseCanaryRecordsProducedFromMetrics( input string) string {
+	regex, _ := regexp.Compile("(?m)^.*strimzi_canary_records_produced_total\\S*\\s(\\d+)$")
+	data := regex.FindStringSubmatch(input)
+	if len(data) > 1 {
+		return data[1]
+	}
+	return ""
+}
+
+func IsTopicPresent(topicName string, topics []string  ) bool{
```
ppatierno, line 77: not used anymore, can be removed

ppatierno (changes requested):

test/e2e_test.go
henryZrncik: this list is empty despite existence of partitions, so I would rather stick with version that worked well.

ppatierno: it's kind of impossible, it's used by the canary ... see `len(topicMetadata.Partitions)` used in the code for example. There should be some other problem is you see it empty.

henryZrncik: Basically you are right, will be done in a minute. 
TLDR
If I create Admin in the beginning of the test, `len(topicsDescriptionList[0].Partitions)` returns 0 whereas former solution 1 (what is correct in default test).  `len(topicsDescriptionList[0].Partitions)` returns correct value as well if I wait some extra time, probably just updated and create admin then. So I can stick to your recommendation, but I will also implement some sort of "waiting".

test/test_utilities.go
```diff
@@ -0,0 +1,92 @@
...
+	regex, _ := regexp.Compile("(?m)^.*strimzi_canary_records_produced_total\\S*\\s(\\d+)$")
+	data := regex.FindStringSubmatch(input)
+	if len(data) > 1 {
+		return data[1]
+	}
+	return ""
+}
+
+func getPartitionCount( clusterAdmin sarama.ClusterAdmin ,topicName string ) int{
+	for i := 0; i < 20; i++ {
```
ppatierno, line 80: why this loop? ... also, you have already called `DescribeTopics` in the function test which calls this function `getPartitionCount` so why doing it again? The topic.go file shows that a describe topics is enough to get the topic and the partitions number.

test/service_manager/service_manager.go
```diff
@@ -0,0 +1,156 @@
...
+func (c *ServiceManager) StartKafkaZookeeperContainers() {
+	log.Println("Starting kafka & Zookeeper")
+
+	var cmd = exec.Command("docker-compose", "-f", c.pathDockerComposeKafkaZookeeper, "up", "-d" )
+	if err := cmd.Run(); err != nil {
+		log.Fatal(err)
+	}
+	log.Println("Zookeeper and Kafka containers created")
+	// after creation of containers we still have to wait for some time before successful communication with Kafka & Zookeper
+	c.waitForBroker()
```
ppatierno, line 50: the canary already has a retry mechanism with exponential backoff for connecting to the Kafka cluster, do we really need another try with another client creation with some attempts and sleep?

ppatierno (changes requested):

scholzj:
Hey @ppatierno @henryZrncik ... Does the PR name really correspond to the content? It looks more like it does add some tests or CI rather than build some new _static_ container.

ppatierno:
@scholzj you are absolutely right. @henryZrncik please could you change the PR name as well? Thanks!

test/test_utilities.go
henryZrncik: because this value needs to be updated, properly. asking too soon ends up in answering with 0 value.

test/service_manager/service_manager.go
henryZrncik: Because I am unable to write some sort of "@BeforeClass" and in some of tests i need to be able to create Sarama Admin (which I am not immediately after creation of Broker). 
In summary waiting has nothing to do with Canary but with need to perform all tests correctly.  Possible solution is move it to tests but as number of those which will require it will be only bigger ...

test/test_utilities.go
henryZrncik: I assume that it is due to asynchronous nature of Kafka, and probably I am asking too soon since topic is created ?

henryZrncik: after second or so provided response match expectation but usually on first attempt empty list is provided.

ppatierno: but in the first DescribeTopics, the topic exists otherwise your test fails correctly ... so it should already have the number of partitions, it doesn't make sense that the admin call reports back metadata for a topic but no info about partitions.

henryZrncik: image

Maybe at least single data produced in that partition I honestly don't know. As I wrote few days later, using old approach it returned correct number immediately.

ppatierno: I will find the time to try/run this and check what Sarama is doing underneath, unless you already debugged it.

henryZrncik: yes I see, I can retry directly here instead of later with second Admin :)

ppatierno: your tests cannot start until the topic is there (no Err) and you know the number of partitions so it would be the first thing to do with just one admin client at the beginning.

Issue #55, opened 2021-05-19
Add SASL mechanisms support for authentication


ppatierno:
It would be great to support different SASL mechanisms for authentication against a Kafka cluster as supported by Strimzi.
The minimum base should be PLAIN.

PR #66, opened 2021-07-06
Added support for SASL PLAIN authentication

ppatierno:
This PR addresses #55

### DIFF
--- a/internal/config/canary_config.go
+++ b/internal/config/canary_config.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Shopify/sarama"
 	"github.com/golang/glog"
 )
 
@@ -36,6 +37,9 @@ const (
 	TLSClientCertEnvVar               = "TLS_CLIENT_CERT"
 	TLSClientKeyEnvVar                = "TLS_CLIENT_KEY"
 	TLSInsecureSkipVerifyEnvVar       = "TLS_INSECURE_SKIP_VERIFY"
+	SASLMechanismEnvVar               = "SASL_MECHANISM"
+	SASLUserEnvVar                    = "SASL_USER"
+	SASLPasswordEnvVar                = "SASL_PASSWORD"
 
 	// default values for environment variables
 	BootstrapServersDefault            = "localhost:9092"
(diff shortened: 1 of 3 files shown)


internal/config/canary_config.go
```diff
@@ -160,10 +173,24 @@ func (c CanaryConfig) String() string {
...
+	SASLPassword := ""
+	if c.SASLMechanism == sarama.SASLTypePlaintext {
+		if c.SASLUser != "" {
+			SASLUser = "[SASL user]"
+		}
+
+		if c.SASLPassword != "" {
+			SASLPassword = "[SASL password]"
+		}
+	}
```
tombentley, line 187: If we just supporting SASL/PLAIN shouldn't we have an `else` here to error on other mechanisms? I can see you handle this in `auth.go`, but it seems cleaner to do it here, no?

ppatierno: this function is just for printing the configuration passed to the canary, it's kind of override of `toString` in Java :-) ... the if condition I am putting there is just for hiding the username/password with placeholders but nothing more.

tombentley: Ah, that's what I get for just scanning the diff. Thanks for the clarification.

tombentley (approved):

k-wall (approved):

internal/security/auth.go
```diff
@@ -25,6 +25,7 @@ func SetAuthConfig(canaryConfig *config.CanaryConfig, saramaConfig *sarama.Confi
 			return errors.New("SASL password must be specified")
 		}
 		saramaConfig.Net.SASL.Enable = true
+		saramaConfig.Net.SASL.Version = sarama.SASLHandshakeV1
```
k-wall, line 28: should that be configurable?

ppatierno: version V0 doesn't provide information about a SASL authentication failure while it's provided by V1 which is supported by ages. I don't see any value supporting V0 imho.

tombentley: See https://cwiki.apache.org/confluence/display/KAFKA/Kafka+APIs

tombentley: Supported since Kafka 1.0 in fact.

PR #79, opened 2021-08-28
Added system certs unit test

ppatierno:
This PR just adds a first unit test for security module when the TLS configuration is created without providing certs but using the system ones.
Other PRs will follow with provided certs.

### DIFF
--- a/internal/security/security_test.go
+++ b/internal/security/security_test.go
@@ -9,6 +9,7 @@
 package security
 
 import (
+	"crypto/x509"
 	"os"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Should it also test when a certificates are passed in the configuration and not just the system store?

ppatierno:
Yes I wrote that in the description to open another PR for that.

scholzj:
@ppatierno Sorry, my mistake ... I complain when _description_ is empty and don't read it when its not :-o

scholzj (approved):

PR #83, opened 2021-09-01
Additional status checks for Kafka operations

MikeEdgar:
This PR adds status checking in two additional places - after describing the Canary's topic and when consuming from the topic.

Signed-off-by: Michael Edgar <medgar@redhat.com>

### DIFF
--- a/internal/services/consumer.go
+++ b/internal/services/consumer.go
@@ -96,10 +96,8 @@ func (cs *ConsumerService) Consume() {
 
 				glog.Infof("Consumer group consume starting...")
 				// this method calls the methods handler on each stage: setup, consume and cleanup
-				err := cs.consumerGroup.Consume(ctx, []string{cs.canaryConfig.Topic}, cgh)
-
-				if err != nil {
-					glog.Infof("Error consuming topic: %v", err.Error())
+				if err := cs.consumerGroup.Consume(ctx, []string{cs.canaryConfig.Topic}, cgh); err != nil {
+					glog.Infof("Error consuming topic: %s", err.Error())
 					return
 				}
 
--- a/internal/services/consumer.go
+++ b/internal/services/consumer.go
@@ -96,7 +96,12 @@ func (cs *ConsumerService) Consume() {
 
 				glog.Infof("Consumer group consume starting...")
 				// this method calls the methods handler on each stage: setup, consume and cleanup
-				cs.consumerGroup.Consume(ctx, []string{cs.canaryConfig.Topic}, cgh)
+				err := cs.consumerGroup.Consume(ctx, []string{cs.canaryConfig.Topic}, cgh)
+
+				if err != nil {
+					glog.Infof("Error consuming topic: %v", err.Error())
+					return
+				}
 
 				// check if context was cancelled, because of forcing a refresh metadata or exiting the consumer
 				if ctx.Err() != nil {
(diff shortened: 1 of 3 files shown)


internal/services/consumer.go
```diff
@@ -96,7 +96,12 @@ func (cs *ConsumerService) Consume() {
 
 				glog.Infof("Consumer group consume starting...")
 				// this method calls the methods handler on each stage: setup, consume and cleanup
-				cs.consumerGroup.Consume(ctx, []string{cs.canaryConfig.Topic}, cgh)
+				err := cs.consumerGroup.Consume(ctx, []string{cs.canaryConfig.Topic}, cgh)
+
+				if err != nil {
+					glog.Infof("Error consuming topic: %v", err.Error())
+					return
+				}
```
ppatierno, line 104: you can use the "short" one line approach here:

```go
if err := cs.consumerGroup.Consume(ctx, []string{cs.canaryConfig.Topic}, cgh); err != nil {
     glog.Infof("Error consuming topic: %v", err.Error())
     return
}
```

Other than this cosmetic change I don't think that "return" is the right way to go. It will exit the go routine and it will never called again (it's part of the Consume that happens on beginning). The error could be transient so maybe we should use a "continue" to try on the next for iteration.

internal/services/topic.go
```diff
@@ -163,8 +162,15 @@ func (ts *TopicService) Reconcile() (TopicReconcileResult, error) {
 		} else {
 			result.Assignments = ts.currentAssignments(topicMetadata)
 		}
-
+	} else {
+		labels := prometheus.Labels{
+			"topic": ts.canaryConfig.Topic,
+		}
+		describeTopicError.With(labels).Inc()
+		glog.Errorf("Error retrieving metadata for topic %s: %v", ts.canaryConfig.Topic, topicMetadata.Err.Error())
```
ppatierno, line 170: `Error()` provides a formatted string so the `%s` instead of `%v` is enough.

ppatierno (changes requested):

internal/services/consumer.go
MikeEdgar: Using "continue" sounds reasonable, but maybe there should be some throttling? The looping hits Kafka very frequently otherwise.

internal/services/topic.go
MikeEdgar: Changed the `%v` to `%s`.

internal/services/consumer.go
ppatierno: I reviewed the code again and as you can see, the overall `Consume()` already has a backoff mechanism based on the fact that after starting the go routing, there is a wait for the consumer to be subscribed (this happens via the `Setup` function signaling on the `ready` channel). If it doesn't happen after a timeout the delay comes into the picture and the go routine is started again. In conclusion, I think that just returning as you are proposing can work. The only minor drawback is waiting for a timeout, even if the consume returned an error but it could be useful for waiting the error, if transient, to be solved. Wdyt?

MikeEdgar: It seems reasonable to me and it's how I have been using this in testing. As you say, the drawback is mixing the error state with a timeout state, but at least it will not repeat the attempts to consume so quickly.

ppatierno (approved):

PR #85, opened 2021-09-08
Added connection check service

### LINKED ISSUE #84: Exposing Kafka cluster connection establishment latency metrics
It would be useful to expose some metrics about the connection establishment latency with the Kafka cluster.
These metrics are also useful when there is a third party authentication service in the picture, so the connection establishment would take into account that part as well (i.e. canary --> Kafka --> authn/z service and back).
The canary could offer a separate "internal" service that periodically opens/closes connections to Kafka brokers checking the latency. This could reveal if something is slowing down with Kafka cluster or the authn part.

ppatierno:
This PR fixes #84.
It adds a new internal service that periodically check connections to the Kafka brokers in the cluster in order to measure latency the reported via corresponding metrics.

### DIFF
--- a/internal/config/canary_config.go
+++ b/internal/config/canary_config.go
@@ -19,101 +19,109 @@ import (
 
 const (
 	// environment variables declaration
-	BootstrapServersEnvVar            = "KAFKA_BOOTSTRAP_SERVERS"
-	BootstrapBackoffMaxAttemptsEnvVar = "KAFKA_BOOTSTRAP_BACKOFF_MAX_ATTEMPTS"
-	BootstrapBackoffScaleEnvVar       = "KAFKA_BOOTSTRAP_BACKOFF_SCALE"
-	TopicEnvVar                       = "TOPIC"
-	ReconcileIntervalEnvVar           = "RECONCILE_INTERVAL_MS"
-	ClientIDEnvVar                    = "CLIENT_ID"
-	ConsumerGroupIDEnvVar             = "CONSUMER_GROUP_ID"
-	ProducerLatencyBucketsEnvVar      = "PRODUCER_LATENCY_BUCKETS"
-	EndToEndLatencyBucketsEnvVar      = "ENDTOEND_LATENCY_BUCKETS"
-	ExpectedClusterSizeEnvVar         = "EXPECTED_CLUSTER_SIZE"
-	KafkaVersionEnvVar                = "KAFKA_VERSION"
-	SaramaLogEnabledEnvVar            = "SARAMA_LOG_ENABLED"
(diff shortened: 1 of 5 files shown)


internal/config/canary_config.go
```diff
@@ -19,101 +19,109 @@ import (
...
-	VerbosityLogLevelEnvVar           = "VERBOSITY_LOG_LEVEL"
-	TLSEnabledEnvVar                  = "TLS_ENABLED"
-	TLSCACertEnvVar                   = "TLS_CA_CERT"
-	TLSClientCertEnvVar               = "TLS_CLIENT_CERT"
-	TLSClientKeyEnvVar                = "TLS_CLIENT_KEY"
-	TLSInsecureSkipVerifyEnvVar       = "TLS_INSECURE_SKIP_VERIFY"
-	SASLMechanismEnvVar               = "SASL_MECHANISM"
-	SASLUserEnvVar                    = "SASL_USER"
-	SASLPasswordEnvVar                = "SASL_PASSWORD"
+	BootstrapServersEnvVar              = "KAFKA_BOOTSTRAP_SERVERS"
```
scholzj, line 22: Should we really use this type of formatting? i'm not sure if it makes the code readable for someone - maybe it does. But it definitely makes PR reviews, diffs and so on much more complicated.

scholzj (approved):
The code looks good to me. I think the formatting sucks, but that is not something to fix in this PR I guess.

internal/config/canary_config.go
ppatierno: I totally agree with you @scholzj but I am just following the Golang best practices where everything is aligned in this way and the tooling you have in the IDE does it for you automatically. This comes from old C ... you know ;-)

scholzj: Should we follow practices which suck? :-o

Anyway, as I said, something to think about ... not really something to change in this PR.

ppatierno: in my opinion it sucks from a PR review point of view and I totally agree with that, but I like this alignment ... maybe because I developed in C a lot in the past? :-)

internal/services/connection_check.go
```diff
@@ -0,0 +1,148 @@
...
+		// ignore error because it will be reported by Connected() call if "not connected"
+		b.Open(cs.client.Config())
+
+		labels := prometheus.Labels{
+			"brokerid": strconv.Itoa(int(b.ID())),
+		}
+
+		if connected, err := b.Connected(); !connected {
+			connectionError.With(labels).Inc()
+			glog.V(1).Infof("Connected broker %d [%t] error [%v]", b.ID(), connected, err)
```
k-wall, line 135: let's log the time taken too.

k-wall (approved):
lgtm

PR #87, opened 2021-09-12
Added util package

ppatierno:
This PR just adds an util package hosting just a common code about getting current time in milliseconds for now.

### DIFF
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -1,0 +1,14 @@
+//
+// Copyright Strimzi authors.
+// License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+//
+
+// Package util contains some utility functions
+package util
+
+import "time"
+
+// NowToMilliseconds returns the current time in milliseconds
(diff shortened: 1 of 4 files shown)


internal/util/util.go
```diff
@@ -0,0 +1,14 @@
...
+// License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+//
+
+// Package util contains some utility functions
+package util
+
+import "time"
+
+// NowToMilliseconds returns the current time in milliseconds
+func NowToMilliseconds() int64 {
```
scholzj, line 12: I wonder if it should be called `NowInMilliseconds()` => you are not converting some passed parameter to milliseconds but instead getting the current time in milliseconds.

scholzj (approved):
One nit, LGTM otherwise.

sknot-rh (approved):
Jakub's comment makes sense to me. Otherwise LGTM

internal/util/util.go
ppatierno: Agree, it makes more sense.

PR #93, opened 2021-09-30
Allowing to configure the canary topic parameters

### LINKED ISSUE #92: Allowing to provide the canary topic configuration
Right now, the canary topic is created with the default configuration related to the Kafka cluster while it would be useful to allow providing the configuration parameters to the canary so that the topic is created using that one.
That could be related, for example, to use a shorter segment size (compared to the default 1 GB) or retention time (compared to the default 7 days) and much more.

ppatierno:
This PR fixes #92 allowing to provide the topic configuration to the canary on creation or for updating the current one when the canary restarts and the topic already exists.


ppatierno:
@k-wall can you have a look as well, please?

internal/services/topic.go
```diff
@@ -139,22 +145,34 @@ func (ts *TopicService) Reconcile() (TopicReconcileResult, error) {
...
+				return result, err
+			}
+		}
+
 		// topic partitions reassignment happens if "dynamic" reassignment is enabled
 		// or the topic service is just starting up with the expected number of brokers
 		if ts.isDynamicReassignmentEnabled() || (!ts.initialized && ts.canaryConfig.ExpectedClusterSize == len(brokers)) {
 
-			glog.Infof("Going to alter topic and reassigning partitions if needed")
+			glog.Infof("Going to reassigning topic partitions if needed")
```
k-wall, line 168: grammar reassigning -> reassign

k-wall (approved):

scholzj (commented):
Kafka is using comma separated lists as values. So how does your code for example handle `cleanup.policy=compact,delete,compression.type=gzip`?

ppatierno:
@scholzj right, good spot ... the current regex `([\w\.]+)=(\w+)` wont get "compact,delete" as value but just "compat" :-( ... I will figure out a solution, thanks!

ppatierno:
@scholzj the list of key-value pairs has to be semicolon separated so that the new regex can handle values containing comma (as per the lists).

scholzj (commented):
Ok, that solves the issue with Kafka values. But it now seems a bit weird with _comma-separated_ list of bootstrap servers and semi-colon separated list of configuration options. Or do you think that is ok?

ppatierno:
I agree, the story was .... 
I started this PR using semicolon separation. Then I worked on the bootstrap servers using the same separator as the Kafka tools which is the comma, then I said "maybe it's better using comma on the topic configuration too to be consistent", so I changed to using comma here ... then you raised the issue, so I came back to semicolon :-)
Using comma for bootstrap servers is consistent with Kafka tooling; using semicolon for topic configuration helps for the lists using comma.

scholzj (approved):
I wonder if we should not rely on the env. vars in the Golang code and instead have just CLI option. In the deployment, you can then pass the env vars into the command line or just configure the command directly. And there you can also do things such as using repeated `--topic-config` option- You can do that when the tool uses env vars internally. But that would require bigger change. so if you wanna merge this as it is, feel free to do so.

ppatierno:
Yeah it would completely change the pattern we are using today. Better merging this as it is for now. Thanks.

PR #96, opened 2021-10-03
Fixed bootstrap servers usage as comma separated list of servers

### LINKED ISSUE #95: KAFKA_BOOTSTRAP_SERVERS list not splitted on commas but used as one server
When the `KAFKA_BOOTSTRAP_SERVERS` env var is set with a comma separated list of kafka brokers, the Sarama client creation fails with following error:

```shell
W1003 09:52:06.827667   15420 main.go:117] Error creating new Sarama client, retrying in 1000 ms: kafka: client has run out of available brokers to talk to (Is your cluster reachable?)
```

This is because the corresponding configuration parameter is used in a wrong way as it would contain just one server:

```go
client, clientErr := sarama.NewClient([]string{canaryConfig.BootstrapServers}, config)
```

It has to be splitted on commas and passed as an array to the Sarama creation function, while right now if the list is i.e. "localhost:9092,localhost:9093", it's used as it was one broker to connect which is obviously not correct.

ppatierno:
This PR fixes #95 using the bootstrap servers parameter as a comma separated list of servers.

### DIFF
--- a/internal/config/canary_config_test.go
+++ b/internal/config/canary_config_test.go
@@ -10,13 +10,15 @@ package config
 
 import (
 	"os"
+	"strings"
 	"testing"
 	"time"
 )
 
 func TestConfigDefault(t *testing.T) {
 	c := NewCanaryConfig()
-	assertStringConfigParameter(c.BootstrapServers, BootstrapServersDefault, t)
+	bootstrapServersDefault := strings.Split(BootstrapServersDefault, ",")
+	assertStringSlicesConfigParameter(c.BootstrapServers, bootstrapServersDefault, t)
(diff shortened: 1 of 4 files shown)


internal/config/canary_config_test.go
```diff
@@ -45,7 +47,7 @@ func TestConfigDefault(t *testing.T) {
 }
 
 func TestConfigCustom(t *testing.T) {
-	os.Setenv(BootstrapServersEnvVar, "my-cluster-kafka:9092")
+	os.Setenv(BootstrapServersEnvVar, "my-cluster-kafka:9092,my-cluster-kafka:9093")
```
scholzj, line 50: This does not seem to make much sense in Strimzi context. Would it be better to have some value like `broker1:9092,broker2:9092`?

scholzj (approved):
One nit, LGTM otherwise.

PR #97, opened 2021-10-03
Strict validation of the topic configuration parameter

ppatierno:
This trivial PR enforces the ";" as separator in the topic configuration parameter because with the current regex, any other special character (i.e. ? or $ or % ...) could be used as separator and being accepted.


scholzj (approved):
Looks good. Maybe some tests with invalid values might be good, but the code logic seems fine.

ppatierno:
Yeah. Just added, thanks!

scholzj (approved):

PR #102, opened 2021-10-05
Fixed log level on consume error

ppatierno:
Trivial PR to use the right error log level on consume function error.

### DIFF
--- a/internal/services/consumer.go
+++ b/internal/services/consumer.go
@@ -100,7 +100,7 @@ func (cs *ConsumerService) Consume() {
 				glog.Infof("Consumer group consume starting...")
 				// this method calls the methods handler on each stage: setup, consume and cleanup
 				if err := cs.consumerGroup.Consume(ctx, []string{cs.canaryConfig.Topic}, cgh); err != nil {
-					glog.Infof("Error consuming topic: %s", err.Error())
+					glog.Errorf("Error consuming topic: %s", err.Error())
 					time.Sleep(consumeDelay)
 					continue
 				}


internal/services/consumer.go
```diff
@@ -100,7 +100,7 @@ func (cs *ConsumerService) Consume() {
 				glog.Infof("Consumer group consume starting...")
 				// this method calls the methods handler on each stage: setup, consume and cleanup
 				if err := cs.consumerGroup.Consume(ctx, []string{cs.canaryConfig.Topic}, cgh); err != nil {
-					glog.Infof("Error consuming topic: %s", err.Error())
+					glog.Errorf("Error consuming topic: %s", err.Error())
```
scholzj, line 103: Should this be maybe just a warning or something when we continue? Or does Golang have no warnings?

ppatierno: Golang has warnings but the `Consume()` function returned an actual error and should be just retried. Do you have concerns about logging as error?

scholzj: I don't know ... we don't fail the canary and exit, we recover from it (through the retry). So in operators we would IMHO use warning.

scholzj: BTW: I can approve it if you want error ... I do not care that much. Just my opinion.

ppatierno: I see your point but it would be better having an overall pass across the other places where a warning could be better. If you don't mind let's approve this one as it is. I will come back with overall review of warnings with a different PR later on.

scholzj (approved):

PR #103, opened 2021-10-08
Add script to calculate canary log latencies and quantiles

racheljpg:
This script gathers the latency values from the canary logs and calculates the average producer/consumer/connection latencies. This may be useful to determine if the histogram buckets for metrics are set optimally.
It takes input from one or more files and also accepts input (as lines) from STDIN. 
It also sorts the information into quantiles. When calculating quantiles, the quantile method and number of cuts are both configurable. If left unspecified, it will use the defaults for these. 
For reference: 
https://docs.python.org/3/library/statistics.html#statistics.quantiles

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,0 +1,45 @@
+# __Canary Latencies Script__
+
+## Table of Contents
+1. Description
+2. Prerequisites
+3. Usage
+4. Resources
+----
+## __Description__
+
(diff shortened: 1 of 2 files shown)


ppatierno:
@scholzj any new comments on this?

scholzj (approved):
@ppatierno TBH, I do not know enough about the canary to understand why is it needed and I'm not sure anyone is curious about my Python comments. So as far as I can review it, it seems ok and if you think this is useful in anyway, feel free to merge it.

PR #104, opened 2021-10-09
Azure pipeline first setup

ppatierno:
First setup of Azure pipeline for running tests and building the binary.


scholzj (approved):
I wonder if we need a script for the `make go_build` command ... but LGTM overall.

scholzj:
I meant that you can do something like this as well I guess:
```
    - bash: "make go_build"
      displayName: Go build
```

But no big deal, we can keep the script as well.

ppatierno:
Ahhhh ... You "we need" had the meaning of the opposite direction, I misunderstood. I agree, let's remove it for now, I was just following the scripts usage in drain cleaner for example. In this case it's pretty straightforward and simple.

PR #111, opened 2021-10-13
Added implementation of canary status information

### LINKED ISSUE #110: Exposing "status" endpoint with Kafka cluster stability
The canary already exposes metrics in Prometheus format.
Anyway, it would be useful to have the canary providing information about "stability" of the Kafka cluster it's interacting with.
By "stability" it means to know what's the percentage of messages produced/consumed in a configurable time window in order to know that the Kafka cluster is functional.
A use case, in the Strimzi land, could be doing a Kafka cluster upgrade in steps (Kafka version, inter broker protocol and so on) but be sure that between steps, the Kafka cluster was rolled and working fine. It's useful when some automation does that, for example.
It could be useful having a kind of `/status` HTTP endpoint on the canary providing such information as just a 200 OK or even adding more info in JSON format as well.

ppatierno:
This PR fixes #110.

It implements a new internal "status" service that could be used for gathering general status information about the canary and exposing them through a dedicated `/status` HTTP endpoint with a JSON object.
Right now, it just provides the percentage of consumed messages in a specific sliding time window (configurable via env var) using a defined sampling rate (configurable via env var) and the JSON is kind of `{ "ConsumedPercentage": 99.56 }`.

NOTE:
One non optimal thing is about tracking the corresponding produced and consumed records value are doubled in the producer and consumer internal services for both the values and the corresponding metrics. This is because of this issue [1] I opened on the Prometheus go client library about allowing to track metric with atomic operation on local variable but missing labels (that we need!)

[1] https://github.com/prometheus/client_golang/issues/923

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ When running the Strimzi canary tool, it is possible to configure different aspe
 
 ## Endpoints
 
-The canary exposes some HTTP endpoints to provide information about status, healthy and metrics.
+The canary exposes some HTTP endpoints to provide information about status, health and metrics.
 
 ### Liveness and readiness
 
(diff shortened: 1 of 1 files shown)


ppatierno:
@tomncooper I reviewed the JSON structure to add the time window and being specific to the consuming part (in the future we could have some status related to different aspects). So right now it's like:

```json
{
  "Consuming": {
    "TimeWindow": 150000, // in ms
    "Percentage": 100
  }
}
```

Of course the `TimeWindow` field is capped to the configured time window or it provides the elapsed time window in real time (when the complete window is not covered yet).

PR #131, opened 2021-11-01
Added support for SASL SCRAM-SHA 256 and 512 authentication

### LINKED ISSUE #130: Add support for SALS mechanisms SCRAM SHA 256 and 512
Currently, the canary supports only PLAIN as SASL mechanism for authentication.
We should add SCRAM-SHA 256 and 512 as well, as supported by the underlying used Sarama library.

ppatierno:
This PR fixes #130 
This PR adds support for SASL SCRAM-SHA 256 and 512 mechanisms.
It uses the scram client inspired by the example in the Sarama library [1].

[1] https://github.com/Shopify/sarama/issues/2055

### DIFF
--- a/internal/security/scram_client.go
+++ b/internal/security/scram_client.go
@@ -1,23 +1,10 @@
-// Copyright (c) 2013 Shopify
+//
+// Copyright Strimzi authors.
+// License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+//
 
-// Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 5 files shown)


internal/security/scram_client.go
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2013 Shopify
```
scholzj, line 1: So is this one-to-one copy? Might be worth adding some comments / links to where it came from.

ppatierno: I changed by inspiring to it with a reference to the example.

scholzj (approved):

PR #135, opened 2021-11-05
Fix #133: add synchronizing token to canary messages

### LINKED ISSUE #133: Long latency recorded by canary in the event of interrupted connection to broker
If the connection to the broker is interrupted for any reason, internally the canary restores the connection (thanks to sarama retry logic).  This is good, but the latency recorded for the probe message then includes the time taken to reconnect.  This gives a distorted view of message latency behaviour and in our case, leads to spurious alerts.

(In our use-case we use connections.max.reauth.ms, so as sarama doesn't handle [KIP-368](https://cwiki.apache.org/confluence/display/KAFKA/KIP-368) we see frequent disconnections, which exacerbates the issue).

I believe it would be better of this condition were counted as a producer error (or if this behaviour can be made configurable).

k-wall:
..to detect messages produced by previous connections (or previous invocation) to avoid skewing message latency results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Added support for SCRAM-SHA-256 and SCRAM-SHA-512 authentication
 * Added Python script tool to calculate latencies and quantiles from canary log for producer, consumer and connection buckets
+* #133: Add synchronizing token to canary messages to detect messages sent by previous invocations of the canary or even previous instances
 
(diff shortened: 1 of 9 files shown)


k-wall:
After reflection, I don't like this idea. It is confused.  The canary's job is simply to measure latency, if event X means that the message is delayed then the latency figure should reflect that.

PR #147, opened 2021-11-22
Increase timeout for broker wait in e2e tests

kornys:
Signed-off-by: David Kornel <kornys@outlook.com>


Frawless (approved):

sknot-rh (approved):

scholzj (commented):
120s seems like a lot. Can you provide more details why is it needed?

kornys:
> 120s seems like a lot. Can you provide more details why is it needed?

Because we test it in slow environment...

ppatierno (approved):

PR #149, opened 2021-11-23
Refresh describe brokers when not the expected numbers in connection

### LINKED ISSUE #141: Connection check service missing latency metrics for one broker
While running an Apache Kafka cluster with the canary alongside it (setting an expected size of 3 brokers), we noticed that the metrics related to the connection check service are measuring latency on 2 out of 3 brokers.

```shell
strimzi_canary_connection_latency_count{brokerid="0",connected="true"} 2102
strimzi_canary_connection_latency_count{brokerid="2",connected="true"} 2103
```

In the example above, it's missing metrics about broker 1.
It doesn't always happen but could be a race condition when not all the 3 brokers are ready up and running and the connection service describes the cluster just on the beginning for getting all brokers info [1].
This happens only when the expected size is fixed, because with the dynamic way it's normal to measure latency only for running brokers and not the missing ones.

[1] https://github.com/strimzi/strimzi-canary/blob/main/internal/services/connection_check.go#L125

ppatierno:
This PR fixes #141.
The problem was that the connection service just asked for the cluster topology once and maybe when a Kafka cluster is starting up not all the brokers are ready.
For this reason, the change is about the connection service to refresh the describe cluster until the "expected" number of brokers is reached. This way at the beginning it just tests the already running brokers but at some point it will start testing the last starting ones; from this point in time the topology is not refreshed anymore and if brokers go out and in, it will just checks failed connections increasing the metrics.
Different with dynamic scaling up and down, it will test only the running brokers (part of the topology) so refreshing the topology every time (this behaviour doesn't change from current status).

### DIFF
--- a/internal/services/connection_check.go
+++ b/internal/services/connection_check.go
@@ -122,7 +122,7 @@ func (cs *ConnectionService) connectionCheck() {
 		cs.admin = admin
 	}
 
-	if cs.isDynamicScalingEnabled() || cs.brokers == nil {
+	if cs.isDynamicScalingEnabled() || cs.canaryConfig.ExpectedClusterSize != len(cs.brokers) {
 		cs.brokers, _, err = cs.admin.DescribeCluster()
 		if err != nil {
 			if err.Error() == "EOF" {


internal/services/connection_check.go
```diff
@@ -122,7 +122,7 @@ func (cs *ConnectionService) connectionCheck() {
 		cs.admin = admin
 	}
 
-	if cs.isDynamicScalingEnabled() || cs.brokers == nil {
+	if cs.isDynamicScalingEnabled() || cs.canaryConfig.ExpectedClusterSize != len(cs.brokers) {
```
scholzj, line 125: Does `len(nil)` return 0?

k-wall (approved):
looks reasonable to me.

internal/services/connection_check.go
k-wall: i had that question, I satisfied myself that if cs.brokers == nil, len(cs.brokers) == 0 https://pkg.go.dev/builtin#len

scholzj (approved):

grdryn (approved):

internal/services/connection_check.go
ppatierno: Yeah, in the end a "slice" in Golang is a struct with three fields: len, cap and a pointer. A `nil` slice has a len = cap = 0 and a pointer = nil and for it the `len` function returns the actual len which is 0 in the struct. An empty slice is different because the pointer is not `nil` but pointing to a backing array which can be empty, so in that case the `len` in the slice is still 0 but it's not a `nil` slice but an empty one. In the end, it's the len field of the slice which makes sense for the function.

PR #151, opened 2021-11-24
Added handling of connection reset by peer error

### LINKED ISSUE #148: canary left inoperative following "connection reset by peer" condition
We are using strimz-canary 0.1.0.  Following a rollout of canary to our fleet of kafka clusters, one canary instance was left inoperative following an  "connection reset by peer" network condition.  After this point, the canary reported "broken pipe" on every reconciliation attempt.

```
I1123 09:22:28.490831       1 canary_manager.go:119] Canary manager reconcile ...
I1123 09:22:29.090849       1 canary_manager.go:129] ... reconcile done
I1123 09:22:33.490871       1 canary_manager.go:119] Canary manager reconcile ...
I1123 09:22:33.874492       1 canary_manager.go:129] ... reconcile done
I1123 09:22:38.490832       1 canary_manager.go:119] Canary manager reconcile ...
I1123 09:22:38.909494       1 canary_manager.go:129] ... reconcile done
I1123 09:22:43.490844       1 canary_manager.go:119] Canary manager reconcile ...
I1123 09:22:43.950559       1 canary_manager.go:129] ... reconcile done
I1123 09:22:48.490873       1 canary_manager.go:119] Canary manager reconcile ...
E1123 09:22:48.492371       1 topic.go:129] Error describing cluster: read tcp 10.131.92.103:49302->35.174.38.5:443: read: connection reset by peer
I1123 09:22:48.492395       1 canary_manager.go:129] ... reconcile done
I1123 09:22:53.490847       1 canary_manager.go:119] Canary manager reconcile ...
E1123 09:22:53.490924       1 topic.go:129] Error describing cluster: write tcp 10.131.92.103:49302->35.174.38.5:443: write: broken pipe
I1123 09:22:53.490938       1 canary_manager.go:129] ... reconcile done
```

this situation continued for 97 minutes, when for reasons unknown the problem resolved itself and the canary began reconciling normally.  There was no intervention.  Neither brokers nor the canary were restarted.

```
1123 11:01:28.490909       1 topic.go:129] Error describing cluster: write tcp 10.131.92.103:49302->35.174.38.5:443: write: broken pipe
I1123 11:01:28.490923       1 canary_manager.go:129] ... reconcile done
I1123 11:01:33.490847       1 canary_manager.go:119] Canary manager reconcile ...
E1123 11:01:33.490897       1 topic.go:129] Error describing cluster: write tcp 10.131.92.103:49302->35.174.38.5:443: write: broken pipe
I1123 11:01:33.490911       1 canary_manager.go:129] ... reconcile done
I1123 11:01:38.490832       1 canary_manager.go:119] Canary manager reconcile ...
I1123 11:01:38.645301       1 canary_manager.go:129] ... reconcile done
I1123 11:01:43.490839       1 canary_manager.go:119] Canary manager reconcile ...
```

[canary.log](https://github.com/strimzi/strimzi-canary/files/7587621/canary.log)

ppatierno:
This PR fixes #148.
Other than checking the "EOF" error, it also adds the check on the "connection reset by peer" when broker close connection that Sarama is still not able to handle by reconnecting.


internal/services/connection_check.go
```diff
@@ -125,7 +126,7 @@ func (cs *ConnectionService) connectionCheck() {
 	if cs.isDynamicScalingEnabled() || cs.canaryConfig.ExpectedClusterSize != len(cs.brokers) {
 		cs.brokers, _, err = cs.admin.DescribeCluster()
 		if err != nil {
-			if err.Error() == "EOF" {
+			if strings.Contains(err.Error(), util.ErrEof) || strings.Contains(err.Error(), util.ErrConnectionResetByPeer) {
```
scholzj, line 129: Is it intentional that you changed the condition from _is EOF_ to _contains EOF_?

internal/util/errors.go
```diff
@@ -0,0 +1,12 @@
+//
+// Copyright Strimzi authors.
+// License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+//
+
+// Package util contains some utility functions
+package util
+
+const (
+	ErrEof                   = "EOF"
```
k-wall, line 10: I don't know sarama, but does it expose the underlying network err codes, so you can do something more idiomatic.  Relying of string comparisons feels weak.

```
errors.Is(err, syscall.ECONNRESET)
```

k-wall:
@ppatierno out of interest, did you find a reasonable way to reproduce the issue?

internal/util/errors.go
```diff
@@ -0,0 +1,12 @@
+//
+// Copyright Strimzi authors.
+// License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+//
+
+// Package util contains some utility functions
```
scholzj, line 6: Is this really a _utility function_? I wouldn't normally complain ... but the comment makes it weird :-o

internal/services/connection_check.go
ppatierno: Just to generalize as the connection reset by peer that is longer.

internal/util/errors.go
ppatierno: I know, the problem is that Sarama seems to mask that error but it's worth trying, maybe I did something bad.

ppatierno: I made a mistake when checking the error so `errors.Is` was not working.
Right now I tried with `errors.Is(err, io.EOF)` because I am able to reproduce it and it works.
See my new changes.

internal/util/errors.go
ppatierno: I removed this file in my latest commit :-)

ppatierno:
@scholzj @k-wall I was using strings because I thought that Sarama was masking underlying errors but I was making a mistake by checking this, so I changed to use a better `errors.Is` by checking on `io.EOF` and `syscall.ECONNRESET`.
While I am able to check that it works for EOF, by shutting down or killing a broker, I am still not able to reproduce a connection reset from a broker :-(

k-wall (approved):

k-wall:
>  I am still not able to reproduce a connection reset from a broker :-(

I haven't tried this but have you seen https://github.com/Shopify/toxiproxy#reset_peer

ppatierno:
Yes I was playing with toxiproxy yesterday evening, I will keep you posted ;-)

k-wall:
One further thought:   I wonder if adding syscall.EPIPE would be wise too?  going back to the original log, the was the second error we saw was an `EPIPE`.  I don't understand networking sufficiently well to know whether there is a possibility we see  an `EPIPE` without a `ECONNRESET`.   Strikes me it is better safe .. I'd advocate for `errors.Is(err, syscall.EPIPE)` too.

E1123 09:22:48.492371       1 topic.go:129] Error describing cluster: read tcp 10.131.92.103:49302->35.174.38.5:443: read: connection reset by peer
I1123 09:22:48.492395       1 canary_manager.go:129] ... reconcile done
I1123 09:22:53.490847       1 canary_manager.go:119] Canary manager reconcile ...
E1123 09:22:53.490924       1 topic.go:129] Error describing cluster: write tcp 10.131.92.103:49302->35.174.38.5:443: write: broken pipe

tombentley:
+1 for `EPIPE`, AFAIK `send()` can result in `EPIPE` when the connection is closed (see https://www.gnu.org/software/libc/manual/html_mono/libc.html#Operation-Error-Signals).

ppatierno:
@k-wall @tombentley I used toxiproxy together with a simple cluster with just one broker in order to have a simple test case with the canary connection going through the toxiproxy and the broker advertising the toxiproxy address.
When I add a toxic `reset_peer` I actually see the RST packet at TCP level via Wireshark but on the canary side it doesn't result in a ECONNRESET every time but quite often in a EOF.
Anyway I was able to reproduce the error a few times and it was recovered by canary successfully.

Screenshot-20211125150723-669x339

ppatierno:
@scholzj do you have any more comments on this. I changed the PR so in the end your comments are outdate.

tombentley:
@ppatierno For `EPIPE` are you sure there's no way the canary can end up trying to send over the socket after the client has been closed? From my extremely scant knowledge that would result in an `EPIPE` and wouldn't be demonstrable via testing with a flaky proxy.

scholzj (approved):

ppatierno:
@tombentley I was able to test the original subject of the PR, the ECONNRESET not the EPIPE. 
Around all these errors (broken pipe included), the canary will catch them and close the connection by itself before establishing a new one so it's not sending on a "broken" channel anymore.
This is the workaround we already had in place for EOF due to Sarama not able to handle reconnection by itself internally.

ppatierno:
@tombentley just to add that the original problem with EOF and now with ECONNRESET is that the canary continues to use the current closed socket so you can see the "broken pipe" error as well in the Keith log (on the issue).
It's not happening anymore if EOF and this PR should save from ECONNRESET as well so that if they happen the channel is closed and a new one is opened. Of course, adding the EPIPE handling saves more catching the specific error if the previous ones don't happen.

tombentley (commented):
Factor out a `isDisconnection()` function, perhaps?

ppatierno:
@tombentley yeah makes sense, I factored out the function.

tombentley (approved):

PR #156, opened 2021-12-01
Test e2e logs update

henryZrncik:
## 🍀 Change
---
Τhis PR adds the logging into e2e tests with the verbose flag (`-v`). Moreover, it adds logging to the executor commands (i.e., compose actions).

### DIFF
--- a/test/service_manager/service_manager.go
+++ b/test/service_manager/service_manager.go
@@ -6,6 +6,7 @@
 package service_manager
 
 import (
+	"bytes"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


test/service_manager/service_manager.go
```diff
@@ -109,12 +118,39 @@ func (c *ServiceManager) waitForBroker() {
...
 	}
 }
 
+func (c *ServiceManager) executeCmdWithLogging( commandDescription ,commandName string, commandArgs ...string) error{
+
+	cmd := exec.Command(commandName, commandArgs...)
+	var stdout, stderr bytes.Buffer
+	// redirect Stdout and Stderr of command to buffers
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
```
ppatierno, line 143: AFAIU we are redirecting entire std out and err in two bytes buffer in memory, are we sure it's going to work? what about the max size of these buffers? I have doubt about this.

ppatierno (approved):
Let's stay with using in memory buffer for now, unless it will show up as a problem in the future.

PR #164, opened 2022-01-25
adding unit test for IsDisconnection

k-wall:
Signed-off-by: kwall <kwall@apache.org>

### DIFF
--- a/internal/util/util_test.go
+++ b/internal/util/util_test.go
@@ -22,6 +22,7 @@ func TestIsDisconnection(t *testing.T) {
 		{errors.New("foobar"), false},
 		{io.EOF, true},
 		{syscall.EPIPE, true},
+		{syscall.ECONNRESET, true},
 		{syscall.ETIMEDOUT, true},
 	}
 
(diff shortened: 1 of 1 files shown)


internal/util/util_test.go
```diff
@@ -0,0 +1,34 @@
...
+func TestIsDisconnection(t *testing.T) {
+	cases := []struct {
+		err error
+		expected bool
+	}{
+		{nil, false},
+		{errors.New("foobar"), false},
+		{io.EOF, true},
+		{syscall.EPIPE, true},
+		{syscall.ETIMEDOUT, true},
```
ppatierno, line 25: do we want to test `syscall.ECONNRESET` as `IsDisconnection` function test as well?

ppatierno (changes requested):

internal/util/util_test.go
k-wall: Done

ppatierno (approved):

scholzj (approved):

PR #165, opened 2022-01-25
treat os.ErrDeadlineExceeded (i/o timeout) as a disconnected condition too

k-wall:
In the end I decided it was best to treat `os.ErrDeadlineExceeded` (i/o timeout) as a disconnected condition.  I was influenced by sarama's own error handling  [consumer.go:818
](https://github.com/Shopify/sarama/blob/712b6bb89287b4767557cc4bf9883148b4979a85/consumer.go#L806) actually closes the broker on any net conditions (`err != nil`).  Having the canary handle this net condition in the same way seemed least worst.  I took the cautious approach and just added `os.ErrDeadlineExceeded` to the existing list of conditions.

### DIFF
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -9,6 +9,7 @@ package util
 import (
 	"errors"
 	"io"
+	"os"
 	"syscall"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


ppatierno:
@k-wall how this PR is related to #164? I see this one adds same test but with `ErrDeadlineExceeded` as well. Should we just close #164 ?

k-wall:
#164 tests the existing functionality, that's why I kept it separate.  I thought you might merge that and then I would rebase this PR.

ppatierno:
ok makes sense. I will do this way if you prefer.

ppatierno (approved):

scholzj (approved):

PR #168, opened 2022-02-03
Fix #166: support the dynamic modification of logging levels

### LINKED ISSUE #166: Allow to change canary logging verbosity and enable/disable Sarama logger at runtime
Currently it's possible to specify the logging verbosity of the canary itself and enabling/disabling the Sarama logger via environment variables.
When an error occurs in a canary instance running, it would be great to increase/enable both at runtime without restarting the canary (because the issue could disappear).
One way could be having an "optional" configuration file for the canary (specifying its path via an env var?) where configuration parameters could be specified and overriding what's defined via env vars.
The canary could then have a periodic internal service watching changes to this file and change internal config accordingly.

k-wall:
My initial approach was to use https://github.com/fsnotify/fsnotify, but supporting kube projected configmaps turns out to be quite tricky.  In the end I hand-rolled my own watcher, that uses a checksum to check for changes.   

I kept the configuration approach generic in case in future there is need to make other configuration items dynamic at runtime.


k-wall:
`MUTABLE_CONFIG_FILE` is rather a horrible name to have facing the user. Wondering about a rename to either DYNAMINC_CONFIG_FILE or maybe even just CONFIG_FILE?

internal/config/canary_config_test.go
```diff
@@ -132,6 +133,15 @@ func TestTopicConfigurationEmptyKeyValuePair(t *testing.T) {
 	t.Errorf("Should have been panicked!")
 }
 
+func TestStringer(t *testing.T) {
```
ppatierno, line 136: where is this used?

internal/config/canary_config.go
```diff
@@ -73,10 +75,18 @@ const (
...
 )
 
+type MutableCanaryConfig struct {
+	SaramaLogEnabled              *bool `json:"saramaLogEnabled"`
+	VerbosityLogLevel             *int  `json:"verbosityLogLevel"`
+}
+
 // CanaryConfig defines the canary tool configuration
 type CanaryConfig struct {
+	*MutableCanaryConfig
```
ppatierno, line 89: Does it need to be a pointer?

ppatierno (commented):
Thanks for the PR @k-wall ... I had a first quick pass and left a couple of questions.
Regarding the name, I would say to use `Dynamic` instead of `Mutable` everywhere.
The same is used in Kafka for example, where config parameters that you can change at runtime are called this way.

README.md
```diff
@@ -47,34 +47,55 @@ If you're using the Strimzi User Operator, the values for these environment vari
...
+
+As mentioned above certain aspects of behaviour can be overridden dynamically at runtime from a JSON configuration file.   If a config file reference is
+provided (`MUTABLE_CONFIG_FILE`), that file will be monitored for changes in content and creation/deletion, with any changes being applied dynamically to the
+Canary's runtime state.  In a kubernetes environment this file could be provided by a projected configmap.
+
+```json
+{
+  "saramaLogEnabled": true,
+  "verbosityLogLevel": 1
+}
```
scholzj, line 97: Would something like YAML be a more natural format then JSON for Strimzi? Most of our stuff is YAML. Or why was JSON chosen?

README.md
```diff
@@ -47,34 +47,55 @@ If you're using the Strimzi User Operator, the values for these environment vari
...
+| `STATUS_TIME_WINDOW_MS` | it defines the sliding time window size (in ms) in which status information are sampled. | `300000` |  |
+| `MUTABLE_CONFIG_FILE` | location of an optional external config file that provides configuration at runtime . | empty |  |
+| `MUTABLE_CONFIG_WATCHER_INTERVAL` | interval that mutable config file is examined for changes in content (in ms)  | `30000` |  |
+
+
+## Configuration file
+
+As mentioned above certain aspects of behaviour can be overridden dynamically at runtime from a JSON configuration file.   If a config file reference is
+provided (`MUTABLE_CONFIG_FILE`), that file will be monitored for changes in content and creation/deletion, with any changes being applied dynamically to the
+Canary's runtime state.  In a kubernetes environment this file could be provided by a projected configmap.
```
scholzj, line 91: Too many spaces between sentences (+ line-breaks in the middle of sentence seem weird).

README.md
```diff
@@ -47,34 +47,55 @@ If you're using the Strimzi User Operator, the values for these environment vari
...
-* `SASL_PASSWORD`: password for SASL authentication against the Kafka cluster when one of `PLAIN`, `SCRAM-SHA-256` or `SCRAM-SHA-512` is used. Default empty.
-* `CONNECTION_CHECK_INTERVAL_MS`: it defines how often the tool has to check the connection with brokers (in ms). Default `120000`.
-* `CONNECTION_CHECK_LATENCY_BUCKETS`: buckets of the histogram related to the broker's connection latency metric (in ms). Default `100,200,400,800,1600`.
-* `STATUS_CHECK_INTERVAL_MS`: it defines how often (in ms) the tool updates internal status information (i.e. percentage of consumed messages) to expose outside on the corresponding HTTP endpoint. Default `30000`.
-* `STATUS_TIME_WINDOW_MS`: it defines the sliding time window size (in ms) in which status information are sampled. Default `300000`
+When running the Strimzi canary tool, it is possible to configure different aspects by using the environment variables listed in the following table.
+
+In addition, certain aspects can be overridden dynamically at runtime from a JSON configuration file.  Where this is possible, a field name is provided in the
+table. The configuration file described in more detail the next section.
```
scholzj, line 53: Too many spaces between sentences. Line-breaks in the middle of sentence seem weird - we normally have sentence per line or paragraph per line.

README.md
```diff
@@ -47,34 +47,55 @@ If you're using the Strimzi User Operator, the values for these environment vari
...
-* `SASL_PASSWORD`: password for SASL authentication against the Kafka cluster when one of `PLAIN`, `SCRAM-SHA-256` or `SCRAM-SHA-512` is used. Default empty.
-* `CONNECTION_CHECK_INTERVAL_MS`: it defines how often the tool has to check the connection with brokers (in ms). Default `120000`.
-* `CONNECTION_CHECK_LATENCY_BUCKETS`: buckets of the histogram related to the broker's connection latency metric (in ms). Default `100,200,400,800,1600`.
-* `STATUS_CHECK_INTERVAL_MS`: it defines how often (in ms) the tool updates internal status information (i.e. percentage of consumed messages) to expose outside on the corresponding HTTP endpoint. Default `30000`.
-* `STATUS_TIME_WINDOW_MS`: it defines the sliding time window size (in ms) in which status information are sampled. Default `300000`
+When running the Strimzi canary tool, it is possible to configure different aspects by using the environment variables listed in the following table.
+
+In addition, certain aspects can be overridden dynamically at runtime from a JSON configuration file.  Where this is possible, a field name is provided in the
```
scholzj, line 52: `Where this is possible, a field name is provided in the table`

It took me a while to understand that this is supposed to refer to the `Config key` column. Maybe we can refer to it more directly in this sentence `... in the table in the Config key column`. Or rename the column to something more self-explanatory such as `Dynamic Configuration field name`?

README.md
```diff
@@ -47,34 +47,55 @@ If you're using the Strimzi User Operator, the values for these environment vari
...
+| `SASL_PASSWORD` | password for SASL authentication against the Kafka cluster when one of `PLAIN`, `SCRAM-SHA-256` or `SCRAM-SHA-512` is used. | empty |  |
+| `CONNECTION_CHECK_INTERVAL_MS` | it defines how often the tool has to check the connection with brokers (in ms). | `120000` |  |
+| `CONNECTION_CHECK_LATENCY_BUCKETS` | buckets of the histogram related to the broker's connection latency metric (in ms). | `100,200,400,800,1600` |  |
+| `STATUS_CHECK_INTERVAL_MS` | it defines how often (in ms) the tool updates internal status information (i.e. percentage of consumed messages) to expose outside on the corresponding HTTP endpoint. | `30000` |  |
+| `STATUS_TIME_WINDOW_MS` | it defines the sliding time window size (in ms) in which status information are sampled. | `300000` |  |
+| `MUTABLE_CONFIG_FILE` | location of an optional external config file that provides configuration at runtime . | empty |  |
+| `MUTABLE_CONFIG_WATCHER_INTERVAL` | interval that mutable config file is examined for changes in content (in ms)  | `30000` |  |
```
scholzj, line 84: Should the description fields start with a capital letter? It seems a bit weird in the table with lowercase.

README.md
```diff
@@ -47,34 +47,55 @@ If you're using the Strimzi User Operator, the values for these environment vari
...
+
+
+## Configuration file
```
scholzj, line 87: So, I guess the values from the config file have higher prio than the environment variables? Or does the canary fail in case of conflict? I think this needs to be clarified.

internal/config/canary_config.go
k-wall: No, it is now changed.

README.md
k-wall: I'll switch it.

scholzj: Maybe wait first what @ppatierno things - to make sure we are all in sync.

README.md
scholzj: > or remove the config file in its entirety

Does that mean that when the file is missing it is just ignored without error?

README.md
k-wall: I've taken your suggestion

README.md
k-wall: @ppatierno - what say you?

k-wall:
I addressed the review comments as separate commits, to help re-review.  Let me know I can I squash whenever.

internal/config/dynamic_config_watcher.go
```diff
@@ -0,0 +1,92 @@
...
+	if c.ticker != nil {
+		c.ticker.Stop()
+		c.ticker = nil
+	}
+}
+
+func readAndHash(filename string) (target *DynamicCanaryConfig, h string, err error) {
+	byteValue, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return
```
ppatierno, line 77: are we sure that this return statement is returning the `err` you got from the `ReadFile`?
The previous line uses `:=` (so declare and assign) so I would expect that the `err` you are checking with `nil` is the local variable but then returning the `err` named return value which is `nil`.
Or this local `err` variable is hiding the return named one and it's returned instead of it? Not sure about it ...

README.md
ppatierno: If we think about the canary running in a container and the config file mounted via a ConfigMap, maybe having a JSON in a field of the data section is more readable than having YAML nested into YAML? But anyway I have no strong opinion, I can live with both.

internal/config/canary_config.go
```diff
@@ -73,10 +75,18 @@ const (
 	ConnectionCheckLatencyBucketsDefault = "100,200,400,800,1600"
 	StatusCheckIntervalDefault           = 30000
 	StatusTimeWindowDefault              = 300000
+	DynamicConfigFileDefault             = ""
+	DynamicConfigWatcherIntervalDefault  = 30000
 )
 
+type DynamicCanaryConfig struct {
+	SaramaLogEnabled              *bool `json:"saramaLogEnabled"`
+	VerbosityLogLevel             *int  `json:"verbosityLogLevel"`
```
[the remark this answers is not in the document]
k-wall: Yes, the use of pointers here is deliberate.  It is possible the configuration file will define only a subset of the fields.  In the case where it is absent I want to fallback to the value provided by the environment variable (or the inbuilt default).

internal/config/dynamic_config_watcher.go
k-wall: Yes.  Golang's shorthand declaration declares only variables that don't already exist, so err on the line 75 is the same as err on line 74. 

I can remove the use of the naked return if you prefer.  Originally `read` and `hash` were two separate short functions.  In the longer function maybe the naked return just obfuscates.

README.md
k-wall: Yeah - it was the kube projected configmap use-case that I was thinking about when I chose yaml.   `kubectl edit` defaults to yaml, and whilst embedding a yaml document within a `data.key` works, it is slightly harder for humans.

The user would need to provide:

```
data:
  config.yml: |-
    ---
    saramaLogEnabled: true
    verbosityLogLevel: 1
```

I suspect many users would get confused and instead try:

```
data:
  config.yml:
    saramaLogEnabled: true
    verbosityLogLevel: 1
```

using JSON avoided the mental gymnastics.

> But anyway I have no strong opinion, I can live with both.

Both?  Do you mean either?   I could write the code to handle both JSON and YAML but .. YAGNI

scholzj: TBH, I do not see that as an issue => but if you both want to stick with JSON, fine with me. I do not see much sense in supporting both => that seems unnecessary.

ppatierno: No no I just meant that I am fine with whatever choice you made. I am more for JSON tbh.

internal/config/dynamic_config_watcher.go
ppatierno: > Yes. Golang's shorthand declaration declares only variables that don't already exist, so err on the line 75 is the same as err on line 74.

Right!

> I can remove the use of the naked return if you prefer.

I am fine tbh.

ppatierno (approved):
LGTM. Thanks for the PR!

PR #169, opened 2022-02-09
Fix data race issue introduced by #166

k-wall:
I didn't notice yesterday that I'd introduced a data-race (reported by `go test --race ./...`) in #166.  Sorry about that.

This first change addresses that.   One of the data races was caused by the use of a value receiver for the #Close() function., which I found really subtle (http://tleyden.github.io/blog/2016/05/19/go-race-detector-gotcha-with-value-receivers/ helped me understand).

The second change addresses thread-safety concerns around the manipulation of sarama.Logger from different go routines without being protected by mutex.  This wasn't reported by `go test --race`.  However I think my proposed change is much easier to reason about in terms of correctness.

The final change ensures that the goroutine beneath the watcher actually ends, rather than being left around.

### DIFF
--- a/internal/config/dynamic_config_watcher.go
+++ b/internal/config/dynamic_config_watcher.go
@@ -64,10 +64,9 @@ func NewDynamicConfigWatcher(canaryConfig *CanaryConfig, applyFunc func(config *
 	return dynamicConfigWatcher, nil
 }
 
-func (c DynamicConfigWatcher) Close()  {
+func (c *DynamicConfigWatcher) Close()  {
 	if c.ticker != nil {
 		c.ticker.Stop()
-		c.ticker = nil
 	}
 }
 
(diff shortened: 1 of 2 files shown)


internal/config/dynamic_config_watcher.go
```diff
@@ -31,12 +35,11 @@ func NewDynamicConfigWatcher(canaryConfig *CanaryConfig, applyFunc func(config *
 			applyFunc(target)
 		}
 
-		dynamicConfigWatcher.ticker = time.NewTicker(canaryConfig.DynamicConfigWatcherInterval * time.Millisecond)
-		quit := make(chan struct{})
 		go func() {
 			for {
+				ticker := time.NewTicker(canaryConfig.DynamicConfigWatcherInterval * time.Millisecond)
```
ppatierno, line 40: shouldn't this ticker creation be out of the `for` loop (still inside the go routine of course)? It sounds like every time you get a ticker event and the corresponding code runs in the select, then back to the for loop a new ticker is created.
It sounds to be wrong to me, am I missing something?

k-wall: you are right, that was my intent.

PR #179, opened 2022-05-24
Add tracing with opentelemetry go sarama wrapper

melchiormoulin:
Adding tracing with [opentelemetry sarama wrapper](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/instrumentation/github.com/Shopify/sarama/otelsarama/example/v0.32.0/instrumentation/github.com/Shopify/sarama/otelsarama) 

Why ? 
Add more telemetry data for debugging a particular message, more convenient than logs.

How ?
By putting this env variable ACTIVATE_TRACING=true

Here is just an example of a console tracing.
If this PR is accepted i can handle a more generic approach than the console exporter.


scholzj (commented):
Stupid question ... what does _console tracing_ mean? Always when i used tracing in the past I was sending the traces for example to Jaeger.

melchiormoulin:
the end goal usually is to use jaeger yes. 
What i meant by console tracing is to have the trace in the log like here https://opentelemetry.io/docs/instrumentation/go/getting-started/#creating-a-console-exporter

scholzj:
Thanks for the explanation. I'm not sure how useful really the tracing into the file is on Kubernetes because it brings issues with storage etc. with it. So I wonder if we should target a full support to be able to ship the logs to Jaeger. 

@ppatierno WDYT?

melchiormoulin:
i making the change to target JAEGER if you are ok to integrate the tracing feature

devguyio:
Adding tracing is definitely valuable, however I think it would be great if the tracing backend is configurable , maybe a `ConfigMap`? Here's how it's done in Knative to support Zipkin and Jaeger:
https://knative.dev/docs/eventing/accessing-traces/#configuring-tracing

ppatierno:
I don't think that tracing to the console is something really useful, also bringing problems related to storage already mentioned by Jakub.
I would be more for tracing with systems like Jaeger. Anyway, if user is not using Jaeger but something different like Zipkin or Prometheus (as mentioned in the link you provided), how this is going to work? Should we make it configurable and pluggable?

devguyio:
> I don't think that tracing to the console is something really useful, also bringing problems related to storage already mentioned by Jakub. I would be more for tracing with systems like Jaeger. Anyway, if user is not using Jaeger but something different like Zipkin or Prometheus (as mentioned in the link you provided), how this is going to work? Should we make it configurable and pluggable?

Jaeger provides a zipkin compatible API, so typically it's a good strategy to target zipkin and it'd work for both

SamBarker (commented):
I certainly think tracing is a good idea for the canary. 

Console tracing works for me as a proof of concept, but would need something else to actually go forward. I think the config-map approach suggested else where is a promising way to go.

cmd/main.go
```diff
@@ -49,7 +70,16 @@ func main() {
 	applyDynamicConfig(&canaryConfig.DynamicCanaryConfig)
 
 	glog.Infof("Starting Strimzi canary tool [%s] with config: %+v", version, canaryConfig)
+	if canaryConfig.ActivateTracing {
```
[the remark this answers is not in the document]
melchiormoulin: not sure how it can it work for example if the config is updated just before span.end

cmd/main.go
```diff
@@ -36,7 +41,23 @@ var (
...
 var saramaLogger = log.New(io.Discard, "[Sarama] ", log.LstdFlags)
+
+func InitTracer() *sdktrace.TracerProvider {
+	exporter, err := stdout.New(stdout.WithPrettyPrint())
+	if err != nil {
+		log.Fatal(err)
+	}
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.AlwaysSample()),
+		sdktrace.WithBatcher(exporter),
```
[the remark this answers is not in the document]
SamBarker: Yeah, it will come from the env vars by default and I think thats quite sensible, I just wanted to have a conversation as about if we would rather be explicit.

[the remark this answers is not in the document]
melchiormoulin: you mean add the [link](https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/jaeger#environment-variables) to the readme ?

cmd/main.go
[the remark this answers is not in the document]
melchiormoulin: do you have some advices on this topic ?

cmd/main.go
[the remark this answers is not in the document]
SamBarker: No I meant something more like https://github.com/open-telemetry/opentelemetry-go/blob/main/example/jaeger/main.go#L54

cmd/main.go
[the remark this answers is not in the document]
SamBarker: In a previous life we defaulted to using a no-op tracer, in java, and then swapped that for a proper tracer implementation if configured. That way we didn't have to think about edge cases in most of the code. 

Does the go OTEL library have an equivalent? That way the config listener would just swap between an active and no-op tracer and nothing else needs to care about the flag status.

cmd/main.go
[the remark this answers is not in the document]
melchiormoulin: oh get it. why not using the default env ? Should i pass an already defined parameter service name ?

cmd/main.go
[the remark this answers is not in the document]
melchiormoulin: indeed you are right there is a noop https://github.com/open-telemetry/opentelemetry-go/blob/19294aab4cd8b52afc34b77403e7447109c9b858/internal/global/trace.go#L18-L21
i modified the code

melchiormoulin:
is all good on this PR ?

internal/config/canary_config.go
```diff
@@ -77,11 +78,15 @@ const (
...
+	ActiveTracingDefault                 = false
 )
 
 type DynamicCanaryConfig struct {
-	SaramaLogEnabled  *bool `json:"saramaLogEnabled"`
-	VerbosityLogLevel *int  `json:"verbosityLogLevel"`
+	SaramaLogEnabled      *bool `json:"saramaLogEnabled"`
+	VerbosityLogLevel     *int  `json:"verbosityLogLevel"`
+	ActivateTracing       *bool `json:"activeTracing"`
+	JaegerExporterTracing *bool `json:"jaegerExporterTracing"` //if true activate jaeger else it is opentelemetry exporter
```
SamBarker, line 88: Is there a particular reason to explicitly support Jaeger at this level? 

I'm thinking `ActivateTracing` is enough and we could enable the `JaegerExporter` if we detect the Jaeger env vars.

cmd/main.go
```diff
@@ -36,7 +44,31 @@ var (
...
 var saramaLogger = log.New(io.Discard, "[Sarama] ", log.LstdFlags)
+
+func InitTracer(JaegerExporterTracing bool) *sdktrace.TracerProvider {
+	resources, _ := resource.New(context.Background(),
+		resource.WithFromEnv(), // pull attributes from OTEL_RESOURCE_ATTRIBUTES and OTEL_SERVICE_NAME environment variables
+		resource.WithProcess(), // This option configures a set of Detectors that discover process information
+	)
+	//Could not use OTEL_TRACES_EXPORTER https://github.com/open-telemetry/opentelemetry-go/issues/2310
+	var exporter sdktrace.SpanExporter
+	if JaegerExporterTracing {
```
SamBarker, line 55: Rather than passing the flag in could we read `OTEL_TRACES_EXPORTER` here and load `jaeger` or `otlptracegrpc` as appropriate based on the value of the env var?  

Obviously it would be nicer if https://github.com/open-telemetry/opentelemetry-go/issues/2310 was implemented, on that note maybe it would be better to have our own var for it to avoid conflicts in the future if #2310 gets traction.

README.md
```diff
@@ -52,36 +52,37 @@ In addition, certain aspects can be overridden dynamically at runtime from a JSO
...
+| `DYNAMIC_CONFIG_WATCHER_INTERVAL` | Interval that dynamic config file is examined for changes in content (in ms)                                                                                                                                                                                                                                                                                                                                                                                          | `30000`                 |  |
+| `ACTIVATE_TRACING` | Activate console tracing                                                                                                                                                                                                                                                                                                                                                                                                                                              | `false`                 |  |
```
SamBarker, line 85: Need to add the Jaeger var here if we keep it.

SamBarker (commented):
On the right track for me. 

My only concern is how best to handle the exporter selection.

internal/config/canary_config.go
melchiormoulin: you're right !

cmd/main.go
```diff
@@ -36,7 +44,31 @@ var (
...
+		exporter, _ = jaeger.New(jaeger.WithAgentEndpoint()) //from env variable https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/jaeger#environment-variables
+	} else {
+		os.Getenv("HOST_IP")
```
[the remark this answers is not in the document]
melchiormoulin: bad thing i'm correcting this

cmd/main.go
```diff
@@ -141,5 +181,8 @@ func applyDynamicConfig(dynamicCanaryConfig *config.DynamicCanaryConfig) {
 	} else {
 		saramaLogger.SetOutput(io.Discard)
 	}
+	if dynamicCanaryConfig.ActivateTracing != nil && *dynamicCanaryConfig.ActivateTracing {
```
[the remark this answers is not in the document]
SamBarker: I think this is useful! 

I think it probably should swap back to the NO-OP tracer if disabled.

README.md
melchiormoulin: i put the env variables needed

internal/config/canary_config.go
melchiormoulin: now it is based on env variables definition

cmd/main.go
melchiormoulin: now if jaeger env variables are defined the exporter is jaeger else it ins otlp

cmd/main.go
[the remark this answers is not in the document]
SamBarker: I'm thinking of the case where the config changes during runtime i.e. it starts on we enable the tracer and then it gets disabled the tracer will remain enabled & exporting.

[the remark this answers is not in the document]
SamBarker: I would expect something using [NoopTracerProvider](https://pkg.go.dev/go.opentelemetry.io/otel/trace#NewNoopTracerProvider ) and passing that to`otel.SetTracerProvider`

cmd/main.go
```diff
@@ -36,7 +44,31 @@ var (
...
+		os.Getenv("HOST_IP")
+		exporter, _ = otlptracegrpc.New(context.Background(), otlptracegrpc.WithInsecure())
+	}
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithResource(resources),
+		sdktrace.WithBatcher(exporter),
+	)
+	otel.SetTracerProvider(tp)
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}))
+	return tp
```
SamBarker, line 67: Do we actually need to return the TracerProvider here?

melchiormoulin: for shutdown yes

cmd/main.go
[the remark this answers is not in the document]
melchiormoulin: done

cmd/main.go
```diff
@@ -81,16 +93,16 @@ func main() {
 	applyDynamicConfig(&canaryConfig.DynamicCanaryConfig)
 
 	glog.Infof("Starting Strimzi canary tool [%s] with config: %+v", version, canaryConfig)
-	if canaryConfig.ActivateTracing {
 
-		tp := InitTracer(canaryConfig.JaegerExporterTracing)
-		defer func() {
+	tp := InitTracerProvider(canaryConfig.JaegerExporterTracing, canaryConfig.OtlpExporterTracing)
+	defer func() {
+		if tp != nil {
```
SamBarker, line 99: Why return `nil` in the no-op provider case? I would expect the no-op provider to accept the shutdown call.

internal/config/canary_config.go
```diff
@@ -91,7 +90,7 @@ const (
 type DynamicCanaryConfig struct {
 	SaramaLogEnabled      *bool `json:"saramaLogEnabled"`
 	VerbosityLogLevel     *int  `json:"verbosityLogLevel"`
-	ActivateTracing       *bool //if Jaeger or OTLP env var are defined we activate Tracing
+	OtlpExporterTracing   *bool //if Jaeger or OTLP env var are defined we activate Tracing
```
SamBarker, line 93: What was your thinking behind re-naming the variable/flag? I thought the point of the change was to allow us to swap between tracing being active and in-active at will without the instrumented code having to be aware of the change. Therefore the old name perfectly encapsulated the intent for me.

cmd/main.go
melchiormoulin: NoOptracer is from "go.opentelemetry.io/otel/trace" package and do not have the shutdown method
NewTracerProvider from  "go.opentelemetry.io/otel/sdk/trace" have a shutdown method.
In the go.opentelemetry.io/otel/sdk/trace package there is no NoOptracer

internal/config/canary_config.go
melchiormoulin: traces are activated if JAEGER or OTLP env are defined. This bool was not needed but i can rollback it as you prefer

cmd/main.go
SamBarker: Ah rubbish. ok thanks for confirming.

internal/config/canary_config.go
SamBarker: I liked the idea of separating tracing being turned on or off from the configuration of the endpoint but that's my personal preference.


@ppatierno do you have a preference?

melchiormoulin: I put the activateTracing bool. the dynamicconfig does not work as we have to get a new producer and a new consumer to update the tracing.

melchiormoulin:
is all good ?

internal/config/canary_config.go
melchiormoulin: please tell me what is needed to validate this PR :)

cmd/main.go
```diff
@@ -36,7 +45,41 @@ var (
 )
 
 var saramaLogger = log.New(io.Discard, "[Sarama] ", log.LstdFlags)
+
+func InitTracerProvider(ActivateTracing bool, JaegerExporterTracing bool, OtlpExporterTracing bool) *sdktrace.TracerProvider {
```
ppatierno, line 49: is it needed to be exported in other packages? I see it's used in main only so lowercase `initTracerProvider` should be ok, or?

README.md
```diff
@@ -52,37 +52,39 @@ In addition, certain aspects can be overridden dynamically at runtime from a JSO
...
+| `OTEL_EXPORTER_JAEGER_ENDPOINT` | Activate Jaeger tracing                                                                                                                                                                                                                                                                                                                                                                                                                           | empty                   |                                  |
+| `ACTIVATE_TRACING` | Activate tracing. When true  OTEL_EXPORTER_OTLP_ENDPOINT or OTEL_EXPORTER_JAEGER_ENDPOINT                                                                                                                                                                                                                                                                                                                                                                                                                            | `false`                   | `activateTracing`                |
```
ppatierno, line 87: why all this big "green" change when you have just added a few env vars at the bottom?

cmd/main.go
```diff
@@ -36,7 +45,41 @@ var (
...
+		)
+		otel.SetTracerProvider(tp)
+		return tp
+
+	}
+	otel.SetTracerProvider(trace.NewNoopTracerProvider())
+	return nil
+}
+
+func exporterTracing(JaegerExporterTracing bool, OtlpExporterTracing bool) sdktrace.SpanExporter {
```
ppatierno, line 69: using booleans sound like if more exporters will be added, we have to change this method signature and getting more parameters. Shouldn't it be much better having just one parameter specifying the exporter to be used? It could be just a string "jaeger" or "otlp" and then the function will create the right exporter accordingly. When adding a new exporter, only the function body will be changed (not in the signature) to create the newly added exporter. In the strimzi bridge we do something like that. We are moving to OTel support and there is a config parameter that you can set to "jaeger" or "otel" and the right exporter will be taken into account.
I see you are building these boolean from the corresponding env vars to be set but it could also returns just a string specifying which exporter we want.

cmd/main.go
```diff
@@ -23,6 +31,7 @@ import (
 	"github.com/strimzi/strimzi-canary/internal/servers"
 	"github.com/strimzi/strimzi-canary/internal/services"
 	"github.com/strimzi/strimzi-canary/internal/workers"
+	_ "go.opentelemetry.io/otel/exporters/jaeger"
```
ppatierno, line 34: isn't there a duplicate above?

cmd/main.go
```diff
@@ -36,7 +45,41 @@ var (
...
+}
+
+func exporterTracing(JaegerExporterTracing bool, OtlpExporterTracing bool) sdktrace.SpanExporter {
+	//Could not use OTEL_TRACES_EXPORTER https://github.com/open-telemetry/opentelemetry-go/issues/2310
+	var exporter sdktrace.SpanExporter
+	//If the env variables needed are defined we set the exporter to Jaeger else it is an opentelemetry exporter
+	if JaegerExporterTracing {
+		exporter, _ = jaeger.New(jaeger.WithAgentEndpoint()) //from env variable https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/jaeger#environment-variables
+	} else if OtlpExporterTracing {
+		exporter, _ = otlptracegrpc.New(context.Background(), otlptracegrpc.WithInsecure()) //from env variable https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlptrace/README.md
```
k-wall, line 76: We should be secure by default, so I question the use `otlptracegrpc.WithInsecure()`.  Perhaps an environment variable should give the user the ability to configure this, if they wish.

cmd/main.go
```diff
@@ -36,7 +45,41 @@ var (
...
+	otel.SetTracerProvider(trace.NewNoopTracerProvider())
+	return nil
+}
+
+func exporterTracing(JaegerExporterTracing bool, OtlpExporterTracing bool) sdktrace.SpanExporter {
+	//Could not use OTEL_TRACES_EXPORTER https://github.com/open-telemetry/opentelemetry-go/issues/2310
+	var exporter sdktrace.SpanExporter
+	//If the env variables needed are defined we set the exporter to Jaeger else it is an opentelemetry exporter
+	if JaegerExporterTracing {
+		exporter, _ = jaeger.New(jaeger.WithAgentEndpoint()) //from env variable https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/jaeger#environment-variables
```
k-wall, line 74: How will the canary user diagnose tracing endpoint connection issues?   I noticed `WithLogger` options which got me thinking that we should surface the ability to see the tracer's own logging.

cmd/main.go
melchiormoulin: done

internal/services/producer.go
```diff
@@ -64,6 +68,7 @@ func NewProducerService(canaryConfig *config.CanaryConfig, client sarama.Client)
 	if err != nil {
 		glog.Fatalf("Error creating the Sarama sync producer: %v", err)
 	}
+	producer = otelsarama.WrapSyncProducer(client.Config(), producer)
```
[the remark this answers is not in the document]
melchiormoulin: yes it works

internal/services/consumer.go
```diff
@@ -104,9 +107,12 @@ func NewConsumerService(canaryConfig *config.CanaryConfig, client sarama.Client)
 func (cs *ConsumerService) Consume() {
 	backoff := NewBackoff(maxConsumeAttempts, 5000*time.Millisecond, MaxDefault)
 	for {
+		var h sarama.ConsumerGroupHandler
 		cgh := &consumerGroupHandler{
 			consumerService: cs,
 		}
+		h = otelsarama.WrapConsumerGroupHandler(cgh)
```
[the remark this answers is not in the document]
melchiormoulin: yes it works

internal/config/canary_config.go
```diff
@@ -113,14 +126,24 @@ type CanaryConfig struct {
...
+	activateTracing := lookupBoolEnv(ActivateTracingEnvVar, ActivateTracingEnvVarDefault)
+
+	jaegerExporterTracing := isJaegerTracingExporter()
+	otlpExporterTracing := isOtlpTracingExporter()
 	dynamicCanaryConfig := DynamicCanaryConfig{
-		SaramaLogEnabled:  &saramaLogEnabled,
-		VerbosityLogLevel: &verbosityLogLevel,
+		SaramaLogEnabled:      &saramaLogEnabled,
+		VerbosityLogLevel:     &verbosityLogLevel,
+		OtlpExporterTracing:   &otlpExporterTracing,
```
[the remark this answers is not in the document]
melchiormoulin: i removed from dynamicconfig as it is not reloaded the consumer and producer

cmd/main.go
melchiormoulin: i can't get glog also there is no withlogger for oltp. Is it really important ?

cmd/main.go
melchiormoulin: by removing otlptracegrpc.WithInsecure() if we put the right env variable this is secure

cmd/main.go
melchiormoulin: i modified it

README.md
melchiormoulin: i updated the readme

README.md
```diff
@@ -126,6 +127,8 @@ The `Consuming` field provides information about the `Percentage` of messages co
 }
 ```
 
+If the time window has not ended, the `/status` endpoint cannot report a percentage of correctly consumed messages. Instead, it returns `Percentage: -1`. The canary also logs `Error processing consumed records percentage: No data samples available in the time window ring`.  In this case, you wait until the time window has ended for the sampling to complete. 
+
```
ppatierno, line 131: why GitHub is reporting this line as a change. It was already here, or?

melchiormoulin:
i can't easily do the DCO with the merge, can you help me ?

ppatierno:
@melchiormoulin I think we are going to make a 0.4.0 release out for the latest important bug fixes merged today. The release candidate could be out tomorrow. This PR still needs another pass from all reviewers and I was wondering if you thought about adding some tests. I think it will make 0.5.0 but it doesn't mean that it will be so late after 0.4.0.

melchiormoulin:
can you just help me for dco ? the `git rebase HEAD~16 --signoff` command totally mess up.

scholzj (approved):
LGTM. Thanks for the PR.

cmd/main.go
```diff
@@ -35,9 +43,42 @@ var (
...
+	)
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}))
+	if exporterType != "" {
+		exporter := exporterTracing(exporterType)
+		tp := sdktrace.NewTracerProvider(
+			sdktrace.WithResource(resources),
+			sdktrace.WithBatcher(exporter),
+		)
+		otel.SetTracerProvider(tp)
+		return tp
```
pavolloffay, line 60: nit: this does not read well - e.g. to have two return statements in this function. 

It would be easier to assign `trace.NewNoopTracerProvider()` to a variable and override it in this block if the exporter type is not empty.

cmd/main.go
```diff
@@ -35,9 +43,42 @@ var (
 	}, nil)
 )
 
+
+func initTracerProvider(exporterType string) *sdktrace.TracerProvider {
+	resources, _ := resource.New(context.Background(),
+		resource.WithFromEnv(), // pull attributes from OTEL_RESOURCE_ATTRIBUTES and OTEL_SERVICE_NAME environment variables
+		resource.WithProcess(), // This option configures a set of Detectors that discover process information
+	)
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}))
```
pavolloffay, line 52: context propagation should be as well configurable via env vars

cmd/main.go
```diff
@@ -5,7 +5,15 @@
 package main
 
 import (
+	"context"
 	"flag"
+	"go.opentelemetry.io/otel"
```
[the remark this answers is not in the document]
melchiormoulin: what do you mean by this ?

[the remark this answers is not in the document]
melchiormoulin: done

cmd/main.go
melchiormoulin: this is not comming from same packages one is sdktrace the other is just trace

ppatierno (commented):
I left a couple of more comments but then we should be ok on my side.
You should fix DCO and check why tests are failing.

melchiormoulin:
@ppatierno seems good now

ppatierno (approved):
@melchiormoulin thank you very much! Sorry for this PR taking so long!

PR #183, opened 2022-06-01
Fixed consumer fetch max wait time due to Sarama potential bug

ppatierno:
While testing RC2, we found that the end to end latency on the canary was increased.
After some investigation (thanks to @k-wall) it turned out that the following change in Sarama is the root cause: https://github.com/Shopify/sarama/pull/2227
It highlights that there could be a potential bug elsewhere in Sarama which has to be addressed.
Meanwhile, the canary can be updated by setting that consumer max time to the same 250 ms (as before in Sarama).

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -36,6 +36,7 @@ var (
 )
 
 var saramaLogger = log.New(io.Discard, "[Sarama] ", log.LstdFlags)
+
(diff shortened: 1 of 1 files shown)


k-wall:
lgtm

tombentley (approved):

Issue #190, opened 2022-06-13
Canary topic seen to ignore rack awareness - 6 brokers cluster across 3 AZs


k-wall:
Testing canary 0.2.0 with a 6 broker kafka cluster deployed to kubernetes with nodes across three AZs, we noticed the canary topic partitions were not respecting the kafka rack awareness feature.

The partition assignment looked like this:

```
./bin/kafka-topics.sh {}topic __redhat_strimzi_canary{-}  -describe --bootstrap-server localhost:9096
Topic: __redhat_strimzi_canary    TopicId: T-cSuR5_Te6nEYMec3xSjg    PartitionCount: 6    ReplicationFactor: 3    Configs: min.insync.replicas=2,cleanup.policy=delete,segment.bytes=16384,retention.ms=600000,message.format.version=3.0-IV1,max.message.bytes=1048588
    Topic: __redhat_strimzi_canary    Partition: 0    Leader: 0    Replicas: 0,1,2    Isr: 0,1,2
    Topic: __redhat_strimzi_canary    Partition: 1    Leader: 1    Replicas: 1,2,3    Isr: 1,2,3
    Topic: __redhat_strimzi_canary    Partition: 2    Leader: 2    Replicas: 2,3,4    Isr: 2,3,4
    Topic: __redhat_strimzi_canary    Partition: 3    Leader: 3    Replicas: 3,4,5    Isr: 3,4,5
    Topic: __redhat_strimzi_canary    Partition: 4    Leader: 4    Replicas: 4,5,0    Isr: 4,5,0
    Topic: __redhat_strimzi_canary    Partition: 5    Leader: 5    Replicas: 5,0,1    Isr: 5,0,1
```
 
For instance partition 2 is on 2,3,4 but brokers 2 and 3 are both in AZ `1b`.

oc logs kafka-instance-kafka-2 | grep -E ^broker.rack
broker.rack=us-east-1b

oc logs kafka-instance-kafka-3 | grep -E ^broker.rack
broker.rack=us-east-1b

The canary topic was created by the canary:

```
I0610 13:03:45.234567       1 topic.go:162] The canary topic __redhat_strimzi_canary was created
I0610 13:03:45.234593       1 consumer.go:135] Waiting consumer group to be up and running
```

The brokers seem to have been up before that.

```
kwall@Oslo kas-installer % oc logs kafka-instance-kafka-0 | grep "Startup complete"
[...]

PR #191, opened 2022-06-13
fix: have requestedassignment respect rack

k-wall:
Proposed fix for #190 .

The change works by, in the case where all brokers are all assigned a rack, it orders the brokers used by the `requestedassignment` function in what KIP-36 calls a [rack alternated broker list.](https://cwiki.apache.org/confluence/display/KAFKA/KIP-36+Rack+aware+replica+assignment#:~:text=The%20rack%20alternated%20list%20will%20be). This list is used by the existing algorithm to create the partition assignments.  If not all brokers have rack information, there's no change to existing behaviour.

This will be applied to both initial assignment and any subsequent reassignment.  If broker pods move between racks for any reason, then the existing partitions will be subject to re-assignment by the existing [alterAssignments](https://github.com/strimzi/strimzi-canary/blob/99accac5694fa61ae81b951cae5f5b718ec52cbc/internal/services/topic.go#L343 ) mechanism

### DIFF
--- a/internal/services/topic.go
+++ b/internal/services/topic.go
@@ -219,10 +219,12 @@ func (ts *TopicService) reconcileTopic() (TopicReconcileResult, error) {
 func (ts *TopicService) Close() {
 	glog.Infof("Closing topic service")
 
-	if err := ts.admin.Close(); err != nil {
-		glog.Fatalf("Error closing the Sarama cluster admin: %v", err)
+	if ts.admin != nil {
+		if err := ts.admin.Close(); err != nil {
+			glog.Fatalf("Error closing the Sarama cluster admin: %v", err)
+		}
+		ts.admin = nil
 	}
-	ts.admin = nil
 	glog.Infof("Topic service closed")
 }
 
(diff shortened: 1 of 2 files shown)


k-wall:
The PR currently has contains a unit test which I believe demonstrates the issue.  I have an idea for a code change that should resolve the issue whilst keeping compatibility for clusters are not AZ based deployments.

internal/services/topic_test.go
```diff
@@ -0,0 +1,176 @@
...
+					found := false
+					for _, expectedBrokerId  := range tt.brokersWithMultipleLeaders {
+						if expectedBrokerId == brokerId {
+							found = true
+							break
+						}
+					}
+
+					if !found {
+						t.Errorf("broker %d is leader more than one partition (%d)", brokerId, count)
```
k-wall, line 95: typo: missing "for"

internal/services/topic.go
```diff
@@ -312,14 +314,55 @@ func (ts *TopicService) requestedAssignments(currentPartitions int, brokers []*s
...
+		}
+	} else {
+		index := 0
+
+		for ;; {
+			again := false
+
+			for _, rackName := range rackNames {
+				brokerList := rackMap[rackName]
+				if len(brokerList) > 0 {
```
ppatierno, line 343: can this really happen? If we are here, I guess all brokers have rack if we are here, or? Maybe the check is not needed?

internal/services/topic.go
```diff
@@ -312,14 +314,55 @@ func (ts *TopicService) requestedAssignments(currentPartitions int, brokers []*s
...
+				rackNames = append(rackNames, broker.Rack())
+			}
+			rackMap[broker.Rack()] = append(rackMap[broker.Rack()], broker)
+		}
+	}
+
+	if len(brokers) != brokersWithRack {
+		if brokersWithRack > 0 {
+			glog.Warningf("Not *all* brokers have rack assignments (%d/%d), topic %s will be created without rack awareness", brokersWithRack, len(brokers), ts.canaryConfig.Topic)
+		}
```
ppatierno, line 334: I was wondering if we could move the ordering of brokers IDs here instead of having that operation outside.
I mean, ordering by ID is not useful if we are going to use rack-awareness but only if we don't use rack-awareness. Wdyt?

internal/services/topic.go
```diff
@@ -312,14 +314,55 @@ func (ts *TopicService) requestedAssignments(currentPartitions int, brokers []*s
...
+					head, rackMap[rackName] = brokerList[0], brokerList[1:]
+					brokers[index] = head
+					index++
+					again = true
+				}
+			}
+
+			if !again {
+				break
+			}
```
ppatierno, line 354: AFAIU this block is going to change the brokers order so that we have one after the other brokers in different racks.
For example, let's assume this scenario:

rack "es1" --> brokers 0,1
rack "es2" --> brokers 3,5
rack "es3" --> brokers 2,4

the final brokers list will be something like this:

```shell
es1 | es2 | es3 | es1 | es2 | es3
-----------------------------------------
  0   |   3   |   2  |   1   |   5   |   4
-----------------------------------------
```

Is my understanding right? Can we add a comment (ASCII art example?) to make it clear?

k-wall: > AFAIU this block is going to change the brokers order so that we have one after the other brokers in different racks.
That's right.

> Is my understanding right? Can we add a comment (ASCII art example?) to make it clear?

Good idea, I'll do it.

internal/services/topic.go
k-wall: You are referring to `sort.Slice`, right?  This is still needed even in the rack case as it will ensure that ordering of brokers within the rack is defined too.  Without this the algorithm won't necessarily generate stable results.

internal/services/topic.go
k-wall: I don't understand, the condition controls the termination condition of the outer loop.

internal/services/topic.go
ppatierno: You are right!

internal/services/topic.go
ppatierno: Never mind. The rackMap is updated in the loop so the brokerList which becomes empty at some point.

ppatierno (approved):
@k-wall LGTM. Thanks!

scholzj (commented):
I think that having a description which describes how is it fixed would not hurt too much :-o.

scholzj:
In particular for example what the limitations are => does it handle only the initial creation? Or also any later updates when the racks change int he broker?

k-wall:
> In particular for example what the limitations are => does it handle only the initial creation? Or also any later updates when the racks change int he broker?

I updated the PR's description.  I hope that answers your questions.

scholzj (approved):
It does answer them, yes. Thanks.

scholzj:
Thanks for the PR.

Issue #188, opened 2022-06-09
Producer and end-to-end latency figures skewed up to Consumer.MaxWaitTime ms


k-wall:
Analysis the producer and end-to-end latency numbers collected by the Canary shows there is a problem with these values.  They are being skewed by the effect of `Consumer.MaxWaitTime`.

Testing the Canary against a single broker with logging turned up and connection establishment latency test and the requirement for re-auth turned off, you see the following pattern.  Notice how the producer latencies grow after each cycle, until value of Consumer.MaxWaitTime is reached, at which point the cycle begins again.

```
kwall@Oslo kas-installer % oc logs -f penguin-canary-5d988b576d-x8dwq | grep "Message sent"
I0608 17:46:57.810235       1 producer.go:100] Message sent: partition=0, offset=1797, duration=2 ms
I0608 17:47:02.847453       1 producer.go:100] Message sent: partition=0, offset=1798, duration=35 ms
I0608 17:47:07.876475       1 producer.go:100] Message sent: partition=0, offset=1799, duration=63 ms
I0608 17:47:12.903987       1 producer.go:100] Message sent: partition=0, offset=1800, duration=90 ms
I0608 17:47:17.932695       1 producer.go:100] Message sent: partition=0, offset=1801, duration=120 ms
I0608 17:47:22.963019       1 producer.go:100] Message sent: partition=0, offset=1802, duration=150 ms
I0608 17:47:27.993141       1 producer.go:100] Message sent: partition=0, offset=1803, duration=180 ms
I0608 17:47:33.048429       1 producer.go:100] Message sent: partition=0, offset=1804, duration=236 ms
I0608 17:47:37.826365       1 producer.go:100] Message sent: partition=0, offset=1805, duration=13 ms
I0608 17:47:42.854491       1 producer.go:100] Message sent: partition=0, offset=1806, duration=42 ms
I0608 17:47:47.880666       1 producer.go:100] Message sent: partition=0, offset=1807, duration=68 ms
I0608 17:47:52.911013       1 producer.go:100] Message sent: partition=0, offset=1808, duration=97 ms
I0608 17:47:57.940001       1 producer.go:100] Message sent: partition=0, offset=1809, duration=127 ms
I0608 17:48:02.965793       1 producer.go:100] Message sent: partition=0, offset=1810, duration=153 ms
I0608 17:48:07.995474       1 producer.go:100] Message sent: partition=0, offset=1811, duration=182 ms
[...]

PR #193, opened 2022-06-23
fix: #188 use separate sarama clients for producer and consumer

k-wall:
Sharing a client meant that the client's await for FetchResponse could skew the latency measurement.  This change means that the producer and consumer services create their own clients.  This will increase the number of connections consumed by the canary on the kafka cluster (by 33%).

Signed-off-by: Keith Wall <kwall@apache.org>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.4.0
 
 * Fixed replicas assignment taking rack configuration into account (#190)
+* Use separate sarama clients for producer/consumer in order to reliably measure message latency (#188)
 
 ## 0.3.0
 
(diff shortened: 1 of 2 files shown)


k-wall:
@ppatierno could you please review?  I still need to verify the latency improvement is as expected.  I'll post charts (before and after) once this is done.

scholzj (approved):

ppatierno:
@k-wall other than posting your results about latency before merging, could you push a line on the CHANGELOG mentioning this change please?

k-wall:
Latency results showing the end to end latency for both Strimzi 0.3.0 and Strimzi main + the PR.  I believe this demonstrates that the change is effective,.  One thing the test did highlight is that the default buckets sizes are no longer appropriate.  

Screenshot 2022-06-26 at 19 47 59

PR #194, opened 2022-06-26
fix: adjust default producer/end-to-end latency buckets.

k-wall:
following #188, the default producer/end-to-end histogram buckets used by
ihe canary are no longer ideal with all of the obseverations falling into
the <=100ms bucket.  This change introduces smaller buckets and drops the
largest ones.

Signed-off-by: kwall <kwall@apache.org>


ppatierno (approved):
LGTM. @k-wall can you resolve the conflict please?

scholzj (approved):

PR #203, opened 2022-08-04
Remove duplicated ports from packaging deployment files

Frawless:
This PR removes duplicated `ports` from packaging deployment examples.


scholzj (approved):
@ppatierno How come this was not introduced into the `./install` folder with the 0.5.0 release? These files should be in general in sync - so I wonder how many other issues we have hidden there.

ppatierno:
@scholzj not sure how this happened sorry. But don't think we have other problems.

ppatierno (approved):

PR #211, opened 2023-01-17
Make the docker binary configurable so that podman can be used

tomncooper:
This PR make the docker binary configurable in the Makefile.docker so that `podman` can be used instead. The default is set to `docker` so no changes are needed to continue using the current behaviour.


ppatierno (approved):
LGTM, my friend!

PR #212, opened 2023-02-13
Add option to set constant labels for all Prometheus metrics

tomncooper:
This PR adds the option to set constant labels to all Prometheus metrics produced by the canary. These are defined via a semi-colon separated list of key=value pairs, passed in as an enVar.

This PR also includes an attempt to make some of the e2e tests less flaky.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ The configuration file described in more detail the next section.
 | `DYNAMIC_CONFIG_FILE` | Location of an optional external config file that provides configuration at runtime. | empty |  |
 | `DYNAMIC_CONFIG_WATCHER_INTERVAL` | Interval that dynamic config file is examined for changes in content (in ms)  | `30000` |  |
 | `EXPORTER_TYPE_TRACING` | Tracing Exporter use. Empty value disable tracing, other possible values are `jaeger` or `otlp`  | `` |  |
-
+| `PROMETHEUS_CONSTANT_LABELS` | A list of semicolon separated `key=value` pairs (i.e. `my-label-1=one;my-label-2=two`) that will be set as constant labels on all prometheus metrics. | empty |  |
 
 ## Dynamic Configuration file
 
--- a/internal/config/canary_config.go
+++ b/internal/config/canary_config.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/Shopify/sarama"
 	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 const (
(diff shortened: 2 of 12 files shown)


scholzj (commented):
Why does this make sense in the code? Isn't it easier to use relabeling when scraping them in Prometheus?

tomncooper:
> Why does this make sense in the code? Isn't it easier to use relabeling when scraping them in Prometheus?

Oh that is absolutely the better option. But there are situations where you don't have access to the Prometheus config or can't easily change it and adding a couple of labels directly in the canary is the path of least resistance.

internal/config/canary_config.go
```diff
@@ -48,7 +49,8 @@ const (
 	DynamicConfigFileEnvVar             = "DYNAMIC_CONFIG_FILE"
 	DynamicConfigWatcherIntervalEnvVar  = "DYNAMIC_CONFIG_WATCHER_INTERVAL"
 	//TODO: This will be removed when Support the OTEL_TRACES_EXPORTER env var is available in the SDK see: https://github.com/open-telemetry/opentelemetry-go/issues/2310
-	ExporterTypeTracing = "EXPORTER_TYPE_TRACING"
+	ExporterTypeTracing           = "EXPORTER_TYPE_TRACING"
+	PrometheusConsantLabelsEnvVar = "PROMETHEUS_CONSTANT_LABELS"
```
ppatierno, line 53: please align with previous assignments

README.md
```diff
@@ -83,7 +83,7 @@ The configuration file described in more detail the next section.
 | `DYNAMIC_CONFIG_FILE` | Location of an optional external config file that provides configuration at runtime. | empty |  |
 | `DYNAMIC_CONFIG_WATCHER_INTERVAL` | Interval that dynamic config file is examined for changes in content (in ms)  | `30000` |  |
 | `EXPORTER_TYPE_TRACING` | Tracing Exporter use. Empty value disable tracing, other possible values are `jaeger` or `otlp`  | `` |  |
-
+| `PROMETHEUS_CONSTANT_LABELS` | A list of semicolon separated `key=value` pairs (i.e. `my-label-1=one;my-label-2=two`) that will be set as constant labels on all prometheus metrics. | empty |  |
```
ppatierno, line 86: we are using colon for separating buckets definition, was wondering if it could make sense to use the same here.

ppatierno (commented):
While I agree with Jakub, anyway this doesn't hurt and the default behaviour is the current one. I left a couple of comments.

internal/config/canary_config.go
tomncooper: So this is actually what go fmt enforces. The comment above breaks the alignment.

README.md
tomncooper: The bucket definitions are using "`,`"  separated lists, while the topic config is using "`;`". As this is config I went with the latter.

scholzj (approved):

internal/config/canary_config.go
ppatierno: Isn't it possible to force the alignment we want?

tomncooper: I don't think so no, go fmt is pretty strict.

ppatierno: at this point, why it was ok before? I don't remember I am using go fmt.

tomncooper:
@ppatierno are we ok to merge this?

ppatierno:
@tomncooper I am still bothered by `go fmt`. It was ok before, why should we accept this different indentation?

internal/config/canary_config.go
tomncooper: Maybe this is due to using a newer version of go fmt (1.19 I think)? If you remove the comment the envars line up with the ones above.

ppatierno: can we move that comment on the right of line 52 after the declaration? It should align then, right?

tomncooper: Yes that should work

ppatierno (approved):
LGTM. Thanks Tom!

Issue #215, opened 2023-03-08
Error-related Metrics starting from 0


leewoobin789:
**Description**: Error related metrics starting from 0 instead of instantiating it when it first occurs.
**Reason**: After deployment, building an alert rules based on increase of these metrics is challenging, as the error metrics are only instantiated under the circumstance of error precedence in the past. After restarting the pod, the problem repeats again.
**Solution Proposal**: Starting from 0 right after deployment.

Thank you in advance

ppatierno:
Hey @leewoobin789 sorry but I can't understand what you are describing, could you make an actual example please?

leewoobin789:
hi @ppatierno 
When I create a prometheus alert rule of "round(increase(strimzi_canary_topic_alter_assignments_error_total[5m]), 1) > 0", it won't trigger the alert due to the fact that "increase" function of PromQL does not correctly compare the difference between current and previous values, when there isn't any previous value. therefore even if error occurred, the first alert won't be generated as its value is 0.
if all error-related metrics are initiated from beginning on (instead of later in runtime) with 0, it would correctly detect the increased values in it.

ppatierno:
I am not a Prometheus and PromQL expert but isn't it possible to define some default for the `increase` function? 
Or could using a Promtheus rule returning 0 or the increase function help here?
Of course we could even initialize the metric at 0, are you willing to contribute to the project and opening a PR for that?

leewoobin789:
@ppatierno As far as I know, there isn't workaround for this.
i am btw. willing to contribute :) I will send a PR asap if it is ok for you.

k-wall:
Oh. Interesting. I wasn't aware of this problem before. I agree with @leewoobin789, the canary's behaviour is wrong.

The prometheus [docs](https://prometheus.io/docs/instrumenting/writing_clientlibs/#counter) state that "Counters MUST start at 0.". The prometheus client library can't predict the set of the labels that the canary is going to use, so it seems reasonable that the canary itself has to cause the 0 value to be recorded.

I found this blog post informative: https://www.section.io/blog/beware-prometheus-counters-that-do-not-begin-at-zero/

leewoobin789:
I close this issue because it should be solved by the next version.

leewoobin789:
I close this issue because it should be solved by the next version.

PR #216, opened 2023-03-08
error-related count metrics initialized with 0

leewoobin789:
Issue: https://github.com/strimzi/strimzi-canary/issues/215

All error-related count metrics except "client_creation_error_total" are initialized with value of 0 to prevent PromQL recognizing first exported metric with error to be initial value.
This way, we can define such prometheus alert rule: `increase(strimzi_canary_topic_alter_configuration_error_total[5m]) > 0`

### DIFF
--- a/internal/services/connection_check.go
+++ b/internal/services/connection_check.go
@@ -159,12 +159,10 @@ func (cs *connectionService) connectionCheck() {
 			"brokerid":  strconv.Itoa(int(b.ID())),
 			"connected": strconv.FormatBool(connected),
 		}
-		// initialize all error-related metrics with starting value of 0
-		if connected {
-			connectionError.With(labels).Add(0)
-		}
+
 		if connected {
 			b.Close()
+			connectionError.With(labels).Add(0)
 			glog.V(1).Infof("Connected to broker %d in %d ms", b.ID(), duration)
 		} else {
 			connectionError.With(labels).Inc()
--- a/internal/services/connection_check.go
+++ b/internal/services/connection_check.go
@@ -36,7 +36,6 @@ type connectionService struct {
 	brokers      []*sarama.Broker
 	stop         chan struct{}
 	syncStop     sync.WaitGroup
-	initialized  bool
 }
 
 // NewConnectionService returns an instance of ConnectionService
@@ -62,7 +61,6 @@ func NewConnectionService(canaryConfig *config.CanaryConfig, saramaConfig *saram
 		canaryConfig: canaryConfig,
 		saramaConfig: saramaConfig,
 		admin:        nil,
-		initialized:  false,
 	}
 	return &cs
 }
@@ -162,7 +160,7 @@ func (cs *connectionService) connectionCheck() {
 			"connected": strconv.FormatBool(connected),
 		}
 		// initialize all error-related metrics with starting value of 0
-		if !cs.initialized {
+		if connected {
 			connectionError.With(labels).Add(0)
 		}
 		if connected {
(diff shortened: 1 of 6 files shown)


leewoobin789:
@ppatierno fyi

ppatierno:
@k-wall can you have also a look on this PR please?

internal/services/connection_check.go
```diff
@@ -159,7 +161,10 @@ func (cs *connectionService) connectionCheck() {
 			"brokerid":  strconv.Itoa(int(b.ID())),
 			"connected": strconv.FormatBool(connected),
 		}
-
+		// initialize all error-related metrics with starting value of 0
+		if !cs.initialized {
```
ppatierno, line 165: I cannot understand why we need an "initialization" check here (so using a boolean flag) instead of initializing the `connectionError` in the `NewConnectionService` when it's created for the first time. What am I missing here?

leewoobin789: @ppatierno it was due to the addition of labels ("brokerid", "connected") that are instantiated after attempting to connect.

ppatierno: I would avoid this logic, what about this ...
When it's connected (so the `if connected`) we can call the `connectionError.With(labels).Add(0)`. It means that if it's the first time and connects, initialize the metric to 0; if it's not the first time, adding 0 doesn't change the previous value. Instead if you have an error, you are on the other condition and `connectionError.With(labels).Inc()` is executed.
Could it work? Wdyt?

k-wall:
Oh.   Interesting.  I wasn't aware of this problem before.   I agree with @leewoobin789, the canary's behaviour is wrong.

The prometheus [docs](https://prometheus.io/docs/instrumenting/writing_clientlibs/#counter) state that "Counters MUST start at 0.".   The prometheus client library can't predict the set of the labels that the canary is going to use, so it seems reasonable that the canary itself has to cause the 0 value to be recorded.

I found this blog post informative: https://www.section.io/blog/beware-prometheus-counters-that-do-not-begin-at-zero/

internal/services/connection_check.go
leewoobin789: you are right. is committed

internal/services/producer.go
```diff
@@ -109,6 +124,10 @@ func (ps *producerService) Send(partitionsAssignments map[int32][]int32) {
 			"clientid":  ps.canaryConfig.ClientID,
 			"partition": strconv.Itoa(i),
 		}
+		// initialize all error-related metrics with starting value of 0
+		if !ps.initialized {
+			recordsProducedFailed.With(labels).Add(0)
+		}
```
ppatierno, line 130: can we apply the same argument I made before for the connection check here?
I mean getting rid of `initialized` and `previousNumPartitions` then, if there is an error on sending you have `recordsProducedFailed.With(labels).Inc()` called. In the `else` branch where the latency is computed, you can just add `recordsProducedFailed.With(labels).Add(0)` which initilized the metric for new partitions or just doing nothing for existing ones.
I guess it should work. Wdyt?

ppatierno:
@leewoobin789 I think we are close! Last request is about adding a line in the CHANGELOG related to the 0.7.0 about `Fixed error-related metrics starting from 0 (#215)` or something similar.

leewoobin789:
@ppatierno Done ✅ thanks a lot for reviewing!

ppatierno (approved):
Approved! Thanks again for the PR!

scholzj (approved):

scholzj:
@k-wall Maybe your review here would make sense as well?

k-wall:
I tested @leewoobin789 changes locally, all lgtm.

ppatierno:
Thanks @k-wall !

Issue #143, opened 2021-11-16
Moving from connection_error_total to connection_total metric


ppatierno:
Currently, the `connection_error_total` metric is exposed and it tracks the total number of errors when testing the connection to the brokers.
Anyway, this metric has a "connected" label which is always set to false but it doesn't make any sense this way because the metric itself is clearly about errors.
The purpose of the metric was to reporting the total number of connections, successfully or not, by using the "connected" label with true or false value.
For this reason, the metric name was a mistake and it doesn't reflect successful connections (which are anyway not tracked, unless the latency ones).
Said that it would be better changing this metric to `connection_total` and using the "connected" label the right way to distinguish between success and failure on connection.
To avoid breaking changes, I would deprecate the `connection_error_total` in the first coming version (0.2.0) while adding the new one `connection_total` and then removing the error one in the next version (0.3.0?).

ppatierno:
What I proposed wasn't done in time for 0.2.0 so at this point the plan is delayed.
The `connection_error_total` will be deprecated in the 0.3.0 while adding the new one `connection_total` and then removing the error one in the next version (0.4.0?).
FYI @k-wall

robobario:
Hi @ppatierno, is this issue dead?

ppatierno:
@robobario no I don't think so, you can work on it if you want. Of course the plan is delayed again so I would deprecate in the next 0.7.0 while adding the new one and removing the old one in 0.8.0.

robobario:
Thanks @ppatierno, I was just checking up since we had a ticket to react. But I find myself looking for tasks now ;) so I've pushed up a PR to deprecate the old and start emitting the new metric.

PR #218, opened 2023-04-20
Add connection_total metric and deprecate connection_error_total

robobario:
First part of replacing connection_error_total with connection_total #143

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,8 @@ In order to check how your Apache Kafka cluster is behaving, the Canary provides
 | `consumer_timeout_join_group_total`     | The total number of consumers not joining the group within the timeout            |
 | `consumer_refresh_metadata_error_total` | Total number of errors while refreshing consumer metadata                         |
 | `records_consumed_latency`              | Records end-to-end latency in milliseconds                                        |
-| `connection_error_total`                | Total number of errors while checking the connection to Kafka brokers             |
+| `connection_error_total`                | *Deprecated - use connection_total* Total number of errors while checking the connection to Kafka brokers|
+| `connection_total`                      | Total of outcomes while checking the connection to Kafka brokers                  |
 | `connection_latency`                    | Latency in milliseconds for established or failed connections                     |
 
 Following an example of metrics output.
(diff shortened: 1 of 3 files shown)


internal/services/connection_check.go
```diff
@@ -160,12 +168,13 @@ func (cs *connectionService) connectionCheck() {
 			"connected": strconv.FormatBool(connected),
 		}
 
+		connection.With(labels).Inc()
```
ppatierno, line 171: because of this https://github.com/strimzi/strimzi-canary/pull/216, a contributor had to initialize the error metric to 0. I think that we should do the same when it happens for the new connection metric, otherwise we'll have the same problem.

ppatierno (changes requested):

ppatierno:
@robobario I left a comment, I would also add a line in the 0.7.0 section of the CHANGELOG to mention the new metric addition and the deprecation of the current one.

internal/services/connection_check.go
robobario: Done, I've made it redundantly add `0` to the connection metrics (new and legacy) as soon as we know the broker id. I thought about keeping some state about which brokers have had metrics initialized and only doing it tactically but that felt like overkill.

ppatierno (approved):
LGTM. Thanks for the contribution!

README.md
```diff
@@ -151,7 +151,8 @@ In order to check how your Apache Kafka cluster is behaving, the Canary provides
 | `consumer_timeout_join_group_total`     | The total number of consumers not joining the group within the timeout            |
 | `consumer_refresh_metadata_error_total` | Total number of errors while refreshing consumer metadata                         |
 | `records_consumed_latency`              | Records end-to-end latency in milliseconds                                        |
-| `connection_error_total`                | Total number of errors while checking the connection to Kafka brokers             |
+| `connection_error_total`                | *Deprecated - use connection_total* Total number of errors while checking the connection to Kafka brokers|
```
PaulRMellor, line 154:
```suggestion
| `connection_error_total`                | (Deprecated - `use connection_total`) Total number of errors while checking the connection to Kafka brokers|
```

scholzj (approved):

README.md
```diff
@@ -151,7 +151,8 @@ In order to check how your Apache Kafka cluster is behaving, the Canary provides
...
+| `connection_total`                      | Total of outcomes while checking the connection to Kafka brokers                  |
```
PaulRMellor, line 155:
```suggestion
| `connection_total`                      | Total number of connections to Kafka brokers                  |
```

PaulRMellor: I wasn't sure what _outcomes_ meant here. Suggestion assumes we're just returning total number of connections.  We could add succeeded or failed too: _Total number of (failed or successful) connections to Kafka brokers_

PaulRMellor (approved):
Great. I made a couple of minor suggestions

README.md
robobario: done (with the backticks around connection_total excluding the use)

PR #227, opened 2024-04-24
chore: Upgraded go version from 1.19 to 1.22

tgib23:
Go 1.19 is not supported version, so better to keep updated.
I've updated Go and ran the new code on my setup for around 20 hours.

It functions, and I see a bit less CPU and similar memory usages (yellow : 0.6.0, green : this PR)
image
image

I've included the update in CHANGELOG as well.

PR #228, opened 2024-04-24
chore: Upgraded go version from 1.19 to 1.22 (repro error)

tgib23:
https://github.com/strimzi/strimzi-canary/pull/227 failed without much clues.
I'd like to check if the same error reproduces with this PR.

### DIFF
--- a/.azure/templates/steps/setup_go.yaml
+++ b/.azure/templates/steps/setup_go.yaml
@@ -2,4 +2,4 @@
 steps:
 - task: GoTool@0
   inputs:
-    version: '1.19.3'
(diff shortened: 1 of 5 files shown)


tgib23:
I created this PR to reproduce the error of https://github.com/strimzi/strimzi-canary/pull/227 and investigate on it, but it seems Github status has been unstable yesterday (it's still slow)
https://www.githubstatus.com/incidents/7txcg0j03kkg

I think it's better to retry building https://github.com/strimzi/strimzi-canary/pull/227 and close this PR now.

PR #231, opened 2024-04-30
docs(mtls): add example in combination with 'KafkaTopic' and 'KafkaUser'

### LINKED ISSUE #230: Provide docs using strimzi-kafka-operator
Hey, I stumbled upon this project and tried to deploy it in combination with the strimzi-kafka-operator.
For this I was a bit curious about creating a `KafkaTopic` and `KafkaUser` with ACLs enabled in my cluster.

Can you please provide some docs explaining this.

sebastiangaiser:
closes #230

### DIFF
--- a/packaging/examples/mtls/mtls.yaml
+++ b/packaging/examples/mtls/mtls.yaml
@@ -1,0 +1,118 @@
+---
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaTopic
+metadata:
+  name: canary
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  partitions: 3
+  replicas: 3
+---
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaUser
+metadata:
+  name: strimzi-canary-client
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  authentication:
+    type: tls
+  authorization:
+    type: simple
+    acls:
+      - resource:
+          type: topic
+          name: canary
+          patternType: literal
+        operations:
+          - Describe
+          - Write
+          - Read
+        host: "*"
+      - resource:
+          type: cluster
+        operation: Alter
+        host: "*"
+      - resource:
+          type: group
+          name: strimzi-canary-group
+          patternType: literal
+        operations:
+          - Read
+        host: "*"
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: strimzi-canary
+  labels:
+    app: strimzi-canary
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: strimzi-canary
+  template:
+    metadata:
+      labels:
+        app: strimzi-canary
+    spec:
+      serviceAccountName: strimzi-canary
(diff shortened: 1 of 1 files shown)


ppatierno (changes requested):

packaging/examples/mtls/mtls.yaml
```diff
@@ -0,0 +1,118 @@
...
+      app: strimzi-canary
+  template:
+    metadata:
+      labels:
+        app: strimzi-canary
+    spec:
+      serviceAccountName: strimzi-canary
+      containers:
+      - name: strimzi-canary
+        image: quay.io/strimzi/canary:0.6.0
```
ppatierno, line 64: under packaging folder tag should be latest
```suggestion
        image: quay.io/strimzi/canary:latest
```

ppatierno (changes requested):

ppatierno (approved):
LGTM. Thanks!

scholzj (commented):
* Maybe we should add some README.md into the folder with some explanation?
* Maybe we should make it in the path a bit more clear that this is about linking it with Strimzi-based Kafka cluster and not as muhc about mTLS as the path suggests? E.g. `packaging/examples/strimzi-tls-auth`?
* It seems disconnected from the release tooling. So if we ever do another release of this, nothing will hapen with this. So I guess the `MAkefile` tasks should be updated to:
    * Include this in the release package
    * Copy it to `/examples`
    * Make sure the Deployment YAML is up to date (I think having a separate deployment is a bit controversial as it would get easily out of date. But not sure how to do it better other thn having just a README.md with YAML snippets.)
    * Update the Deployment YAML during release with the release version (to use the correct container)

I guess having this just as a section in the main README.md would safe a lot of these things. But I'm fine if we want to have it in examples. We just have to make sure it is done properly or it will be obsolete and forgotten soon.

sebastiangaiser:
If you want I can move the files to a different path, split them up, ... Please tell me where :D

The structure of the readme and the makefile parts are not on me, I guess. So maybe this should be in another PR.

scholzj:
> If you want I can move the files to a different path, split them up, ... Please tell me where :D
> 
> The structure of the readme and the makefile parts are not on me, I guess. So maybe this should be in another PR.

I think there are several problems with this:
* You do not really have an example. You have installation files. 
    * Installation files need to be properly maintained (e.g. if you add a new environment variable you have to add it to all instances).
    * Installation files need to be package as part of the release
    * Installation files need to be updated as part of the release (if nothing else, the image tag changes)

If we merge it as it is, it will never be part of any release, most people will never find it and it will eventually get outdated until it is one day deleted. That is what I'm afraid of.

sebastiangaiser:
Would the releasing be easier with a Helm chart? I think this would not be that kind of documentation but in case when you run `helm template...` you would also get the "raw" manifests, too.

scholzj:
> Would the releasing be easier with a Helm chart? I think this would not be that kind of documentation but in case when you run helm template... you would also get the "raw" manifests, too.

Well, yes and not I guess. We would still need to integrate it into the tooling and CIs.

